PR #2, opened 2015-01-14
Add the logger implementation, with tests.

nomis52:


nomis52:
@peternewman  @joshua-moyerman 

I think the Logger is done.

src/logger.c
```diff
@@ -0,0 +1,127 @@
...
+
+#include "constants.h"
+
+#include <math.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+LoggerData g_logger;
+
+#ifndef min
```
peternewman: I thought IRC said you didn't need this as it exists in math.h? If you do need it, can you move it out to another file so it can be used elsewhere in the code.

src/logger.h
```diff
@@ -58,12 +61,14 @@ extern inline bool Logging_DataPending() {
...
 extern inline bool Logging_HasOverflowed() {
-  return g_logger_data.overflow;
+  return g_logger.overflow;
 }
 
 /**
  * @brief Log a message.
  * @param str The string to log.
+ *
+ * It's not safe to call Logging_Log from an ISR.
```
peternewman: @note

peternewman:
LGTM

joshua-moyerman:
I don't see modifications being made to the project files. If these are to be built and included on the PIC code, you may want to push the changes to the project file. Otherwise, looks good.

src/logger.c
nomis52: Only on the embedded platform.  The ordering causes a bit of trouble here, since it needs to be after all the includes. If I find myself using it anywhere else I'll move it into a header.

src/logger.h
nomis52: Done

src/logger.c
peternewman: Sorry, it only exists as a macro on the embedded platform?

nomis52: Yes, the pic libraries are slightly different from the normal libc libraries.

PR #4, opened 2015-01-16
Start getting Travis working, fix up some issues that appeared in doing so

peternewman:


peternewman:
Probably needs some input from you here @nomis52 before it's finished. Also feel free to contribute the actual contents of tests/config/ that should be in the repo... :)

.travis.yml
```diff
@@ -0,0 +1,69 @@
+language: cpp
+
+script:
+ - "bash -ex .travis-ci.sh"
+
+matrix:
+  fast_finish: true
+  include:
+    - compiler: clang
```
nomis52: We can drop the double compiler check. Since this will only be run on developer machines, we have far less combinations to tests.

.travis.yml
```diff
@@ -0,0 +1,69 @@
...
+ - if [ "$TASK" = "doxygen" ]; then sudo apt-get install -qq doxygen; fi
+
+after_success:
+  - if [ "$TASK" = "coverity" ]; then cat /home/travis/build/${TRAVIS_REPO_SLUG}/cov-int/scm_log.txt; fi
+  - if [ "$TASK" = "coverage" ]; then coveralls --gcov /usr/bin/gcov-4.8 -b . -E '.*Test\.cpp$' --gcov-options '\-lp' > /dev/null; fi
+
+notifications:
+  irc:
+    channels:
+     - "chat.freenode.net#openlighting"
```
nomis52: Can you comment this out for now. The project is still in stealth mode.

tests/scripts/enforce_licence.py
```diff
@@ -0,0 +1,264 @@
...
+  'common/rdm/testdata/duplicate_pid_value.proto',
+  'common/rdm/testdata/inconsistent_pid.proto',
+  'common/rdm/testdata/invalid_esta_pid.proto',
+  'common/rdm/testdata/test_pids.proto',
+  'common/rdm/testdata/pids/pids1.proto',
+  'common/rdm/testdata/pids/pids2.proto',
+  'examples/ola-dmxconsole.cpp',
+  'examples/ola-dmxmonitor.cpp',
+  'include/ola/gen_callbacks.py',
+  'olad/www/mobile.js',
```
nomis52: We can remove all? of these I think

nomis52:
Thanks for doing this.

You're right to add ax_prog_doxygen.m4.

.travis.yml
peternewman: Done, although I find the idea of a stealth open source project rather a contradiction. :)

tests/scripts/enforce_licence.py
peternewman: Good spot, done.

peternewman:
No worries, was there anything else static apart from ax_prog_doxygen.m4 in your config folder?

.travis.yml
peternewman: I was going to leave it actually, because between clang and gcc we get a better selection of testing; we sometimes get failures on one and not the other and users may be running either.

peternewman:
Looks like we need to merge this in, or get someone to push to trunk before I can sort out Coverity.

.travis.yml
nomis52: Makes sense.

nomis52:
I'll merge as is and then get to work fixing the errors.

peternewman:
Okay, I don't know if things like lint need locking down to only certain directories in ja-rule @nomis52 , potentially likewise Doxygen too.

nomis52:
Yeah, I'll sort that out.

PR #5, opened 2015-01-16
Use a #define to avoid the runtime function pointer, while..,

nomis52:
still allowing us to inject a function for testing.


nomis52:
@joshua-moyerman @peternewman

nomis52:
@nighttrune

nomis52:
@nightrune

src/logger.h
```diff
@@ -34,19 +34,24 @@ extern "C" {
...
- * @param enabled true to enable the logger, false to disable.
+ *   Logging_SendResponse() is called. This can be overridden, see the note
+ *   below.
  * @param max_payload_size The maximum size of the payload to be passed to the
  *   TXFunction. Must be at least 2.
+ *
+ * @note If PIPELINE_TRANSPORT_TX is defined in system_pipeline.h, the macro
+ * will override the tx_cb argument.
+ *
+ * A Logger starts off in the disabled state.
```
peternewman: Does Doxygen work if we pass in two @note's, like two @returns?

peternewman:
LGTM

src/logger.h
nomis52: Maybe , but I think we should be using less @note rather than more. 'If everything is important, then nothing is.'

peternewman: "Multiple adjacent \note commands will be joined into a single paragraph. Each note description will start on a new line."

If you don't want it as a note, can you move "A Logger starts off in the disabled state." up to below the @brief with a blank line in between, so it shows in the extended info, or use @post "Starts a paragraph where the postcondition of an entity can be described."

nomis52: I've moved up the "A Logger starts off ... " and removed @note.

PR #6, opened 2015-01-19
Fix the build

nomis52:


tests/Doxyfile
```diff
@@ -673,7 +673,7 @@ INPUT_ENCODING         = UTF-8
 # *.hxx *.hpp *.h++ *.idl *.odl *.cs *.php *.php3 *.inc *.m *.mm *.dox *.py
 # *.f90 *.f *.for *.vhd *.vhdl
 
-FILE_PATTERNS          = *.cpp *.c *.h *.dox
+FILE_PATTERNS          = *.h *.dox
```
peternewman: Wouldn't it be better to exclude the bits we don't care about? Or is this to force stuff to be documented in the .h?

tests/Makefile.am
```diff
@@ -33,7 +34,8 @@ src_liblogger_la_CFLAGS = $(BUILD_FLAGS)
 # TESTS
 ################################################
 TESTING_FLAGS = -I system_config -I ../src -I include \
-                $(WARNING_FLAGS) $(CPPUNIT_CFLAGS)
+                -std=c++11 \
```
peternewman: Do we have to do this, for some of us still working on some rather old machines (but still in LTS)?

peternewman:
Looking closer now.

tests/Makefile.am
nomis52: I think this will work with g++ 4.6 (precise) . It Certainly works in 4.8 which is in trusty.  Debian Squeeze comes with 4.8

peternewman: Some people are still on 4.4 :). More curious what features of the latest version you're using?

peternewman:
Are you going to fix or exclude the lint too?

tests/Doxyfile
nomis52: The latter for now. Due to the heavy use of inlining I don't yet have a good sense how things will split across the .h and .c files.

PR #9, opened 2015-01-24
Clean up and test the Stream Decoder

nomis52:


nomis52:
@nightrune @joshua-moyerman 

Here is the next bit

src/stream_decoder.c
```diff
@@ -0,0 +1,142 @@
+/*
+ * File:   stream_decoder.h
```
joshua-moyerman: .c*

src/stream_decoder.h
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * @brief A function pointer used to handle new messages.
+ * @param message The message to handle.
+ */
+typedef void (*MessageHandler)(const Message*);
+
+
+/**
+ * @brief Initialize the Stream Decoder.
+ * @param handler The MessageHandler to call with each new message.
```
peternewman: Can we move the param below the other text please.

src/stream_decoder.h
```diff
@@ -0,0 +1,83 @@
...
+bool StreamDecoder_GetFragmentedFrameFlag();
+
+/**
+ * @brief Clear the fragmented frame flag.
+ */
+void StreamDecoder_ClearFragmentedFrameFlag();
+
+/**
+ * @brief Decode data from an input stream.
+ * @param data A pointer to the incoming data.
```
peternewman: Again.

tests/lib/TestUtils.cpp
```diff
@@ -26,7 +26,7 @@
 
 namespace testing {
 
-void _AssertDataEquals(
+void _AssertArrayEquals(
     const CPPUNIT_NS::SourceLine &source_line,
```
peternewman: Are you not bothering to typedef here?

src/stream_decoder.h
nomis52: Any particular reason? I find it easier with the important details at the top and then the extended description below.

tests/lib/TestUtils.cpp
nomis52: Not yet. I'm still trying to decide how to unit test all of this. I'm going to experiment with a third unit testing framework so this will probably all change soon anyway.

src/stream_decoder.h
peternewman: It just seems nicer to me when the input maps more to the output doxygen, so all the info, then the params.

PR #13, opened 2015-02-08
Add a configure check to ensure we have std=c++11 support

peternewman:


peternewman:
@nomis52 this currently sets -std=c++11 in CXXFLAGS. Do we want that across everything, or just in some bits (in which case I can modify the code to not set CXXFLAGS)?

nomis52:
I think it's safe to set -std=c++11 in CXXFLAGS.  Just try it and see.

peternewman:
It's what it's currently doing and Travis is working. Do we bother testing the bits that are explicitly specifying c++11 in their Makefiles though?

nomis52:
The bits that specify c++11 are the tests, so yes, running make check is the way to test this works. 
LGTM

peternewman:
So do you want it removed from the mocks that explicitly specify it?

nomis52:
Yeah, if we're setting it at the top level we don't need it in either of the Makefile.mk .

peternewman:
Done.

nomis52:
LGTM

PR #19, opened 2015-02-10
Fix the coverity defects.

nomis52:


tests/mocks/Matchers.cpp
```diff
@@ -60,23 +52,26 @@ bool DataMatcher::InternalMatchAndExplain(
...
+         << (expected == actual ? " == " : " != ")
+         << "0x" << static_cast<int>(actual) << " ("
+         << ((expected >= '!' && expected <= '~') ?
+             static_cast<char>(expected) : ' ')
+         << (expected == actual ? " == " : " != ")
+         << (actual >= '!' && actual <= '~' ? static_cast<char>(actual) : ' ')
+         << ")";
 
-    matched &= expected == actual;
+      matched &= expected == actual;
```
peternewman: Can we have some brackets please? :)

tests/mocks/MessageHandlerMock.cpp
```diff
@@ -49,22 +49,26 @@ bool MessageMatcher::MatchAndExplain(
...
-       << "\n" << std::dec << i << ": 0x" << std::hex
-       << static_cast<int>(expected)
-       << (expected == actual ? " == " : " != ")
-       << "0x" << static_cast<int>(actual) << " ("
-       << ((expected >= '!' && expected <= '~') ? static_cast<char>(expected) :
-           ' ')
-       << (expected == actual ? " == " : " != ")
-       << (actual >= '!' && actual <= '~' ? static_cast<char>(actual) : ' ')
-       << ")";
+      *listener
```
peternewman: Can we put this in a TestUtils file somewhere rather than duplicating it across two files?

peternewman:
Looks good generally.

tests/mocks/Matchers.cpp
nomis52: Done

tests/mocks/MessageHandlerMock.cpp
nomis52: I considered it. I want to see how this shakes out a bit more before splitting it into a shared library.

PR #20, opened 2015-02-10
Fix Microchip Harmony Config errors

jmoyerman:
Moving firmware files to firmware dir and modifying tests to suit.


nomis52:
LGTM. Merge when the build passes.

PR #36, opened 2015-04-14
Work around the way we run Doxygen so we don't end up with unneccesary a...

peternewman:
...bsolute paths in the generated docs


peternewman:
Another retrospective @nomis52

nomis52:
LGTM

PR #38, opened 2015-04-16
Add a minimum Doxygen version check adapted from https://github.com/mlos...

peternewman:
...kot/autotools-modules/blob/master/macros/ax_prog_doxygen.m4


peternewman:
I guess we may be better off using m4_version_compare.

peternewman:
This is as good as this will get unless I cherry-pick your tweaks to Travis

peternewman:
Actually I need to fix another bug too when Doxygen isn't installed.

peternewman:
Now fixed I think.

PR #41, opened 2015-04-18
Clean up the command documentation some more.

nomis52:


peternewman:
Fix Travis.

firmware/src/constants.h
```diff
@@ -88,25 +88,75 @@
  * @brief The Ja Rule message commands.
  */
 typedef enum {
-  ECHO = 0x80,  //!< Echo the data back
-  TX_DMX = 0x81,  //!< Transmit a DMX frame
+  ECHO = 0x80,  //!< Echo the data back. See @ref message_cmd_echo
+
```
peternewman: Any reason for the space?

firmware/src/constants.h
```diff
@@ -88,25 +88,75 @@
...
+  TX_DMX = 0x81,  //!< Transmit a DMX frame. See @ref message_cmd_txdmx.
   GET_LOG = 0x82,  //!< Fetch more log data
   GET_FLAGS = 0x83,  //!< Get the flags state
   WRITE_LOG = 0x84,
+
+  /**
+   * @brief Reset the device.
+   * See @ref message_cmd_reset.
```
peternewman: I think I asked about @see before?

tests/doxygen/message-format.dox
```diff
@@ -46,7 +46,7 @@
  * @param Token A token for the command. The same token will be returned in the
  * command response. Typically the host will increment the token with each
  * request.
- * @param Command The command ID.
+ * @param Command The \ref Command identifier.
```
peternewman: Any reason for these not being @ref?

peternewman:
LGTM

firmware/src/constants.h
nomis52: The enums are one per line. If I add @see it breaks the second sentence out into a new paragraph and the table gets much bigger.

tests/doxygen/message-format.dox
nomis52: Switched.

firmware/src/constants.h
nomis52: Sorry which space?

firmware/src/constants.h
peternewman: Okay, fair enough.

firmware/src/constants.h
peternewman: The random new space after echo. They don't seem to be grouped into logical groups, just randomly.

nomis52: Oh, yeah, that's coming in a follow up . I'm going to group the enums and set the correct values.

PR #50, opened 2015-05-27
Handle the case where a very short low precedes the break.

nomis52:


firmware/src/constants.h
```diff
@@ -261,6 +261,10 @@ typedef enum {
  *
  * Measured in 10ths of a microsecond. The value is from line 2 of Table 3-1
  * in E1.20.
+ *
+ * Note that some responders, like say the Robin 600 Spot, have a very short low
```
peternewman: Add the manufacturer name?

firmware/src/transceiver.c
```diff
@@ -310,17 +310,11 @@ void __ISR(_INPUT_CAPTURE_2_VECTOR, ipl6) InputCaptureEvent(void) {
...
         if (g_timing.get_set_response.mark_start -
             g_timing.get_set_response.break_start <
             CONTROLLER_RX_BREAK_TIME_MIN) {
-          // The break was too short.
-          g_transceiver.result = T_RESULT_RX_INVALID;
-          PLIB_TMR_Stop(TMR_ID_3);
-          Transceiver_ResetToMark();
-          g_transceiver.state = STATE_COMPLETE;
+          // The break was too short, go back to STATE_RX_WAIT_FOR_BREAK
+          g_transceiver.state = STATE_RX_WAIT_FOR_BREAK;
```
peternewman: Log a warning?

nomis52: It's within an ISR so you can't log. It's also within the spec so isn't a problem now that we handle it.

firmware/src/constants.h
nomis52: It tried to keep it reasonably obtuse but still enough info that I can remember how to reproduce it :)

peternewman: Okay

firmware/src/transceiver.c
peternewman: Understood

peternewman:
LGTM

PR #58, opened 2015-06-17
Stub out the mode change command

nomis52:
It doesn't do anything yet.

### DIFF
--- a/firmware/src/app.c
+++ b/firmware/src/app.c
@@ -24,12 +24,15 @@
 #include "coarse_timer.h"
 #include "logger.h"
 #include "message_handler.h"
+#include "rdm.h"
 #include "stream_decoder.h"
 #include "syslog.h"
 #include "system_definitions.h"
 #include "system_settings.h"
 #include "transceiver.h"
 #include "usb_transport.h"
+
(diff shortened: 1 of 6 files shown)


firmware/src/app.c
```diff
@@ -31,6 +32,8 @@
 #include "transceiver.h"
 #include "usb_transport.h"
 
+const uint8_t OUR_UID[UID_LENGTH] = {0x7a, 0x70, 0xff, 0xff, 0xfe, 0};
```
peternewman: This should probably have a todo to move to a new range when Ja Rule gets a proper block allocated.

firmware/src/constants.h
```diff
@@ -114,6 +114,12 @@ typedef enum {
    */
   COMMAND_RESET_DEVICE = 0x00,
 
+  /**
+   * @brief Change the operating mode of the device.
+   * See @ref message_cmd_set_mode.
```
peternewman: Leave a blank line or @sa

peternewman:
LGTM; fix Travis.

firmware/src/app.c
nomis52: Added in the next commit

nomis52:
Fixed

PR #64, opened 2015-07-11
Add the dot diagrams for the controller and responder.

nomis52:


peternewman:
LGTM

You should update https://github.com/OpenLightingProject/ja-rule/blob/master/tests/config/ax_prog_doxygen.m4#L265 and Travis.

peternewman:
And do the same as here:
https://github.com/OpenLightingProject/ola/pull/708

peternewman:
We need to install graphviz via Travis too @nomis52 .

nomis52:
We are:

 - if [ "$TASK" = "doxygen" ]; then sudo apt-get install -qq doxygen graphviz; fi

peternewman:
Ah yes, I'd done it previously.

PR #68, opened 2015-07-12
Clean up the timing parameters

nomis52:


firmware/src/transceiver.h
```diff
@@ -379,7 +379,7 @@ bool Transceiver_SetRDMDUBResponseLimit(uint16_t limit);
 
 /**
  * @brief Return the Controller DUB response timeout.
- * @returns The RDM DUB response limit, in 10ths of a millisecond.
+ * @returns The RDM DUB response limit, in 10ths of a microsecond.
```
peternewman: Ah, this was kind of why I was asking the question before.

firmware/src/transceiver_timing.h
```diff
@@ -0,0 +1,162 @@
...
+ * @file transceiver_timing.h
+ */
+
+// Common params
+// ----------------------------------------------------------------------------
+
+/**
+ * @brief The minimum break time the user can configure, in microseconds.
+ *
+ * DMX1990 was 88uS and later versions raised this to 92uS. We allow down to
```
peternewman: All these s's for seconds should be lowercase (the u should be a mu, but we'll let that slide)

peternewman:
Just some minor comments.

PR #71, opened 2015-07-13
Add a command to fetch the UID of the device.

nomis52:


firmware/src/message_handler.c
```diff
@@ -73,6 +75,19 @@ static void SetMode(uint8_t token,
   SendMessage(token, COMMAND_SET_MODE, RC_OK, NULL, 0);
 }
 
+static void GetUID(uint8_t token, unsigned int length) {
+  if (length) {
+    SendMessage(token, COMMAND_SET_MODE, RC_BAD_PARAM, NULL, 0);
```
peternewman: Should this be COMMAND_GET_MODE?

peternewman: Or even COMMAND_GET_UID

peternewman:
Just one comment

firmware/src/message_handler.c
nomis52: Fixed.

PR #72, opened 2015-07-14
Add the SPI RGB code

nomis52:


nomis52:
Should be ready for review now. This is just basic SPI support for 2 pixels. I'll work on hooking up the RDM personalities / manufacturer PIDs in the next change.

firmware/src/app.c
```diff
@@ -80,14 +81,26 @@ void APP_Initialize(void) {
...
   Logger_Initialize(NULL, PAYLOAD_SIZE);
   Logger_SetState(true);
+
+  SPIRGBConfiguration spi_config;
+  spi_config.module_id = SPI_ID_1;
+  spi_config.baud_rate = 1000000;
+  spi_config.use_enhanced_buffering = true;
+  SPIRGB_Init(&spi_config);
+
+  // Send a frame with all pixels set to 0.
```
peternewman: Is that a good idea/necessary? You'll then get a flash to black if you repower it.

firmware/src/responder.c
```diff
@@ -200,6 +207,13 @@ void Responder_Receive(const TransceiverEvent *event) {
         }
         break;
       case STATE_DMX_DATA:
+        // TODO(simon): configure this with DMX_START_ADDRESS and footprints.
+        if (g_offset - 1 < 6) {
```
peternewman: The 6 is for two pixels yes? Can we at least make it 3*2 for now?

firmware/src/spi_rgb.c
```diff
@@ -0,0 +1,115 @@
...
+ * spi_rgb.c
+ * Copyright (C) 2015 Simon Newton
+ */
+
+#include "spi_rgb.h"
+
+#include "peripheral/spi/plib_spi.h"
+#include "syslog.h"
+
+// TODO(simon): move these into the config (and set with RDM?)
```
peternewman: Yes to the RDM bit. We should define a few manufacturer PIDs which are common between this and OLA. I also think we should split the protocol from the personality. Pixel count and protocol seems like the only stuff we'll need to set via RDM, slots per pixel and latch bytes can be inferred from the protocol.

firmware/src/spi_rgb.c
```diff
@@ -0,0 +1,115 @@
...
+#define PIXEL_COUNT 2
+#define SLOT_PER_PIXEL 3
```
peternewman: SPaG SLOTS_PER_PIXEL

firmware/src/spi_rgb.c
```diff
@@ -0,0 +1,115 @@
...
+  g_spi.use_enhanced_buffering = config->use_enhanced_buffering;
+  g_spi.in_update = false;
+  g_spi.tx_index = 0;
+
+  unsigned int i = 0;
+  for (; i < SLOT_PER_PIXEL * PIXEL_COUNT; i++) {
+    g_spi.pixels[i] = LPD8806_PIXEL_BYTE;
+  }
+
+  for (i = SLOT_PER_PIXEL * PIXEL_COUNT;
```
peternewman: Can we not do memset?

firmware/src/responder.c
```diff
@@ -200,6 +207,13 @@ void Responder_Receive(const TransceiverEvent *event) {
...
+          SPIRGB_SetPixel((g_offset - 1) / 3, (g_offset - 1) % 3, b);
```
peternewman: Does this need a cast to an int?

tests/mocks/SPIRGBMock.cpp
```diff
@@ -0,0 +1,59 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Library General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * SPIRGBMock.cpp
+ * A mock coarse timer module.
```
peternewman: Correct

tests/mocks/SPIRGBMock.h
```diff
@@ -0,0 +1,38 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Library General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * SPIRGBMock.h
+ * A mock coarse timer module.
```
peternewman: Correct

peternewman:
Generally LGTM, just a few comments

firmware/src/app.c
nomis52: Yeah, the SPI bus seems to spew random data on boot, so you get multi-colored rainbows. Resetting to black seems the most sensible thing to do.

firmware/src/responder.c
nomis52: Already fixed in the next change :)

firmware/src/spi_rgb.c
nomis52: Yeah, I was trying to decide if we do personalities or manufacturer PIDs. The downside of manufacturer PIDs is that we can't yet set them in the OLA web UI. But they are much more powerful that trying to shoehorn personalities, so I think I'll go that direction.

firmware/src/spi_rgb.c
nomis52: Fixed

firmware/src/spi_rgb.c
nomis52: fixed

firmware/src/responder.c
nomis52: Nope

tests/mocks/SPIRGBMock.cpp
nomis52: Fixed

tests/mocks/SPIRGBMock.h
nomis52: Fixed

peternewman:
LGTM

firmware/src/app.c
peternewman: Understood.

firmware/src/responder.c
peternewman: Okay

firmware/src/spi_rgb.c
peternewman: We can easily add manufacturer PIDs to the OLA web UI though. We should try and standardise on a PID across the two codebases anyway.

PR #75, opened 2015-07-16
Add macros.h

nomis52:


firmware/src/macros.h
```diff
@@ -0,0 +1,40 @@
...
+#ifndef FIRMWARE_SRC_MACROS_H_
+#define FIRMWARE_SRC_MACROS_H_
+
+/**
+ * @def UNUSED
+ * @brief Mark unused arguments & types.
+ *
+ * @examplepara
+ *   @code
+ *   void Foo(OLA_UNUSED int bar) {}
```
peternewman: Fix the example

firmware/src/macros.h
```diff
@@ -0,0 +1,40 @@
...
+ *
+ *   OLA_UNUSED typedef int Baz;
```
peternewman: Again

peternewman:
LGTM when comments fixed

firmware/src/macros.h
nomis52: Done

PR #77, opened 2015-07-17
Switch to using RDMHandler as the entry point for RDM responders.

nomis52:


firmware/src/rdm_handler.c
```diff
@@ -130,6 +130,14 @@ void RDMHandler_HandleRequest(const RDMHeader *header,
   }
 }
 
+void RDMHandler_GetUID(uint8_t *uid) {
+  if (g_rdm_handler.active_model) {
```
peternewman: Also check it's not the NULL model?

firmware/src/rdm_handler.h
```diff
@@ -104,6 +104,14 @@ void RDMHandler_HandleRequest(const RDMHeader *header,
                               const uint8_t *param_data);
 
 /**
+ * @brief Get the UID of the responder.
+ * @param uid A pointer to copy the UID to; should be at least UID_LENGTH.
```
peternewman: [out]

firmware/src/rdm_handler.c
nomis52: NULL model is an int, this is a pointer to the active model.

peternewman: Understood

firmware/src/rdm_handler.h
nomis52: Done

peternewman:
LGTM

peternewman:
LGTM

PR #80, opened 2015-07-17
Updating project configuration

jmoyerman:


nomis52:
LGTM

peternewman:
Do we need to adapt the tarball checker?

nomis52:
I don't plan on doing tarballs for this, since it's not something we plan to distribute to end users.

Unfortunately the build files here are netbeans XML, unless someone wants to write an XML parser .....

nomis52:
LGTM, does this fix the build?

peternewman:
I wasn't actually suggesting a tarball, just the same concept.

We could use an existing XML parser, or I suspect grep and sed would be enough to get the relevant filenames out and compare them to a tree walk.

PR #81, opened 2015-07-18
Add the simple RDM model

nomis52:


nomis52:
This hooks it all together.

firmware/src/rdm_responder.c
```diff
@@ -269,6 +275,62 @@ int RDMResponder_GetSupportedParameters(const RDMHeader *header,
...
+    .dmx_footprint = htons(g_responder.dmx_footprint),
+    .current_personality = g_responder.current_personality,
+    .personality_count = g_responder.personality_count,
+    .dmx_start_address = htons(g_responder.dmx_start_address),
+    .sub_device_count = htons(g_responder.sub_device_count),
+    .sensor_count = g_responder.sensor_count
+  };
+
+  /*
+  // TODO make this .foo format
```
peternewman: Haven't you done this above?

nomis52: Yup, removed

firmware/src/simple_model.h
```diff
@@ -0,0 +1,76 @@
...
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ *
+ * simple_model.h
+ * Copyright (C) 2015 Simon Newton
+ */
+
+/**
+ * @defgroup simple_model Simple RDM Model
+ * @brief A simple RDM Model for controlling RGB pixels.
```
peternewman: The simple one isn't doing RGB pixels (yet). Is it worth the simple one just being a straight dummy one anyway.

nomis52: Here's what I'm thinking:
 - 'Simple' (RGB controller)
 - Moving light
 - Sub-device dimmer
 - Sensor only

Each will use the SPI pixels (well, with the exception of sensor only) but the 'simple' one is sort of a native SPI RGB controller device.

tests/tests/RDMResponderTest.cpp
```diff
@@ -34,16 +35,45 @@
 #include "MessageHandlerMock.h"
 
 using ola::rdm::UID;
+/*
```
peternewman: Remove

peternewman:
Generally LGTM, just need the fixes.

firmware/src/simple_model.h
peternewman: I think we should have a straight/clean SPI device too though, like the SPI plugin, then one with more features as the simple one. So people can use it for genuine SPI control if they want, without the baggage of fake language setting etc.

tests/tests/RDMResponderTest.cpp
nomis52: Done

peternewman: Where @nomis52 ?

PR #82, opened 2015-07-20
Add the MODEL_ID & MODEL_ID_LIST parameters

nomis52:


nomis52:
Ok, this is ready to go.

firmware/src/macros.h
```diff
@@ -35,4 +35,13 @@
 #define UNUSED
 #endif
 
+/**
+ * @brief A helper to return if the request does not merit a response.
+ * @param header The RDMHeader of the request.
+ */
+#define ReturnUnlessUnicast(header) \
```
peternewman: Didn't we have this already too?

nomis52: It was in rdm_responder.c, I moved it here so I can reuse it for all the models.

firmware/src/rdm.h
```diff
@@ -265,6 +265,14 @@ typedef enum {
 
 
 /**
+ * @brief Open Lighting Project manufacturer PIDs.
+ */
+typedef enum {
+  PID_DEVICE_MODEL = 0x8002,
+  PID_DEVICE_MODEL_LIST = 0x8003
+} OpenLightingManufacturerPID;
```
peternewman: Add an alternative DEVICE_MODEL_DESCRIPTION/LABEL to get descriptions without having to switch into the mode.

firmware/src/rdm.h
```diff
@@ -265,6 +265,14 @@ typedef enum {
 
 
 /**
+ * @brief Open Lighting Project manufacturer PIDs.
+ */
+typedef enum {
+  PID_DEVICE_MODEL = 0x8002,
```
peternewman: I know we've failed so far, but I wonder if we should try and follow this from RDM 6.2.10.2 for these and future PIDs "Manufacturer-Specific PID values should be selected by choosing the appropriate category from Table A-3 and adding an offset of 0x8000 to preserve a logical organization."

firmware/src/rdm.h
nomis52: I don't think it worth it since this is just a testing device. We can add them later but I'm trying to minimize the number of PIDs that are intercepted by the rdm_handler.

firmware/src/rdm.h
nomis52: I really don't like that part of the standard. I don't know anyone who follows it and it's not even obviously from Table A-3 what the category prefixes are.

firmware/src/rdm_handler.c
```diff
@@ -41,6 +43,116 @@ typedef struct {
 
 static RDMHandlerState g_rdm_handler;
 
+static int GetSetModelId(const RDMHeader *header,
```
peternewman: We can do the same to the mute commands now. :)

firmware/src/rdm_handler.c
```diff
@@ -41,6 +43,116 @@ typedef struct {
...
+    g_rdm_buffer[sizeof(RDMHeader) + 1] = ShortLSB(model_id);
+    return RDMUtil_AppendChecksum(g_rdm_buffer);
+
+  } else if (header->command_class == SET_COMMAND) {
+    if (header->param_data_length != sizeof(uint16_t)) {
+      return RDMResponder_BuildNack(header, NR_FORMAT_ERROR);
+    }
+
+    // take action
+    uint16_t new_model = (param_data[0] << 8) + param_data[1];
```
peternewman: JoinUInt8

firmware/src/rdm_handler.c
```diff
@@ -41,6 +43,116 @@ typedef struct {
...
+
+    if (RDMUtil_UIDCompare(our_uid, header->dest_uid)) {
+      return RDM_RESPONDER_NO_RESPONSE;
+    }
+
+    if (!ok) {
+      uint16_t nack_reason = htons(NR_DATA_OUT_OF_RANGE);
+      RDMResponder_BuildHeader(
+          header, NACK_REASON,
+          (header->command_class == GET_COMMAND ?
```
peternewman: This is only SET now.

firmware/src/rdm_handler.c
nomis52: Not quite, this exists outside of the dispatching function, so we don't test it with function pointer comparisons.

firmware/src/rdm_responder.c
```diff
@@ -132,14 +116,14 @@ int RDMResponder_HandleDUBRequest(const uint8_t *param_data,
 void RDMResponder_BuildHeader(const RDMHeader *incoming_header,
                               RDMResponseType response_type,
                               RDMCommandClass command_class,
-                              RDMPid pid,
+                              uint16_t pid,
```
peternewman: Do we need to change this, is it because our manufacturer PIDs aren't in the enum?

firmware/src/rdm_handler.c
nomis52: Done

firmware/src/rdm_responder.c
nomis52: Yup.

firmware/src/macros.h
peternewman: Cool

firmware/src/rdm.h
peternewman: Okay

firmware/src/rdm_handler.c
peternewman: Ah okay

peternewman:
LGTM

PR #84, opened 2015-07-21
Clean up the docs, add some more enums

nomis52:


firmware/src/rdm_handler.h
```diff
@@ -21,8 +21,14 @@
...
  *
- * This represents the entry point for RDM request handling. The handler
- * dispaches the request to the appropriate model.
+ * The RDM Handler manages dispatching the RDM requests to the correct model.
+ * This allows the physical device to simulate different types of RDM devices
+ * from LED drivers and moving lights to sensor-only devices and fog machines.
+ *
+ * Only a single model can be active at once, the PID_DEVICE_MODEL parameter
+ * can be used to change models at run time. Note that PID_DEVICE_MODEL and
+ * PID_DEVICE_MODEL_LIST aren't included in SUPPORTED_PARAMETERS so there are
```
peternewman: SPaG they are

peternewman: If we hide them, we won't be able to easily include them in the OLA web UI to simplify switching.

nomis52: Correct, but people will be able to switch from the console anyway, which is preferred solution. Or you can use the OLA CLI.

firmware/src/rdm_model.h
```diff
@@ -63,11 +64,19 @@ typedef enum {
...
+   * @brief A responder that contains only sensors.
+   */
+  SENSOR_MODEL_ID = 0x0103
 } ResponderModel;
 
 /**
  * @brief Model ioctl enums.
+ *
+ * Ioctls are the generic 'catch-all' operations. We use them so we can reduce
+ * the number of functions we need to implement each models.
```
peternewman: SPaG implement for each model | implement each model

peternewman:
Just minor comments, LGTM

firmware/src/rdm_model.h
nomis52: Done

firmware/src/rdm_handler.h
peternewman: I thought you were potentially aiming it at console manufacturers too and hence would want to make it easy for them to switch and test. I guess we could do something cheeky in OLA and show it based on matching the manufacturer and a particular subset of model IDs. If it's not in on the webpage, then the description of the model before switching becomes irrelevant anyway.

nomis52: It is aimed at console manufacturers but that's why the out of band access is important. You don't want to be using the system you're trying to test to configure the test rig.

peternewman: Fair point, but we could simplify it for others by allowing OLA to configure it too.

peternewman:
LGTM

PR #90, opened 2015-07-23
Add the network responder

nomis52:


nomis52:
I want to move a couple of the constants over to rdm.h but otherwise this should be good.

peternewman:
Fix Travis

firmware/src/network_model.c
```diff
@@ -0,0 +1,837 @@
...
+#include "rdm_buffer.h"
+#include "rdm_frame.h"
+#include "rdm_responder.h"
+#include "rdm_util.h"
+#include "utils.h"
+
+// Various constants
+#define SOFTWARE_VERSION 0x00000000
+
+enum { ETHERNET_HARDWARE_TYPE = 1};
```
peternewman: Keep all the generic ones centrally (so just not NUMBER_OF_INTERFACES).

firmware/src/network_model.c
```diff
@@ -0,0 +1,837 @@
...
+enum { IPSEC_HARDWARE_TYPE = 31 };
+enum { NUMBER_OF_NAMESERVERS = 3 };
+enum { NUMBER_OF_INTERFACES = 3 };
+enum { DNS_HOST_NAME_SIZE = 63 };
+enum { DNS_DOMAIN_NAME_SIZE = 231 };
+enum { MAC_SIZE = 6 };
+enum { INTERFACE_ID_SIZE = 4 };
+
+typedef enum {
```
peternewman: Move these centrally

firmware/src/network_model.c
```diff
@@ -0,0 +1,837 @@
...
+  DHCP_MODE_INACTIVE = 0x00,
+  DHCP_MODE_ACTIVE = 0x01,
+  DHCP_MODE_UNKNOWN = 0x02
+} DHCPMode;
+
+typedef enum {
```
peternewman: Again

firmware/src/network_model.c
```diff
@@ -0,0 +1,837 @@
...
+  IPSEC_INTERFACE_ID = 3,
+  WLAN_INTERFACE_ID = 4
+} InterfaceId;
+
+static const char DEVICE_MODEL_DESCRIPTION[] = "Ja Rule Network Device";
+static const char SOFTWARE_LABEL[] = "Alpha";
+static const char DEFAULT_DEVICE_LABEL[] = "Ja Rule";
+static const char DEFAULT_HOSTNAME[] = "responder";
+static const char DEFAULT_DOMAINNAME[] = "local";
+static const uint32_t NO_DEFAULT_ROUTE = 0;
```
peternewman: And these, apart from DHCP_FAILURE_RATIO.

firmware/src/network_model.c
```diff
@@ -0,0 +1,837 @@
...
+   .id = IPSEC_INTERFACE_ID,
+   .hardware_type = IPSEC_HARDWARE_TYPE,
+   // No h/w address for ptp links
+   .hardware_address = {0, 0, 0, 0, 0, 0},
+  },
+  {
+   .label = WIFI_INTERFACE_NAME,
+   .id = WLAN_INTERFACE_ID,
+   .hardware_type = ETHERNET_HARDWARE_TYPE,
+   // Local admin MAC address
```
peternewman: Correct the comment

firmware/src/network_model.c
```diff
@@ -0,0 +1,837 @@
...
+
+/*
+ * @brief Check a request that contains an interface id.
+ * @returns -1 if the id was found and index was set, otherwise the size of the
+ * RDM response that was created.
+ */
+static int CheckInterfaceRequest(const RDMHeader *header,
+                                 const uint8_t *param_data,
+                                 unsigned int *index) {
+  if (header->param_data_length != sizeof(uint32_t)) {
```
peternewman: Shouldn't this be >= for set with an ID and a parameter.

nomis52: Then it allows too much.  This function is only used for GETs.

As an aside, I have a better way of doing this which is coming in the PR after this one.

firmware/src/network_model.c
nomis52: I've put all of the values from the E1.37-2 doc into rdm.h

firmware/src/network_model.c
nomis52: Done

firmware/src/network_model.c
```diff
@@ -0,0 +1,837 @@
...
+  if (header->param_data_length) {
+    return RDMResponder_BuildNack(header, NR_FORMAT_ERROR);
+  }
+
+  ReturnUnlessUnicast(header);
+
+  unsigned int offset = sizeof(RDMHeader);
+  unsigned int i = 0;
+  for (; i < g_network_model.interface_count; i++) {
+    g_rdm_buffer[offset++] = UInt32Byte0(INTERFACE_DEFINITIONS[i].id);
```
peternewman: If this is actually in the right order, shouldn't the functions be named more logically, so 3 is MSB

nomis52: it's the same order as arrays : [0, 1, 2, 3]

firmware/src/network_model.c
```diff
@@ -0,0 +1,837 @@
...
+  int index = LookupIndex(id);
+  if (index < 0) {
+    return RDMResponder_BuildNack(header, NR_DATA_OUT_OF_RANGE);
+  }
+
+  if (index == IPSEC_INTERFACE_ID) {
+    return RDMResponder_BuildNack(header, NR_ACTION_NOT_SUPPORTED);
+  }
+
+  g_network_model.interfaces[index].configured_zeroconf_mode = param_data[4];
```
peternewman: Replace 4 with sizeof(uint32_t).

firmware/src/network_model.c
```diff
@@ -0,0 +1,837 @@
...
+    return r;
+  }
+
+  ReturnUnlessUnicast(header);
+  InterfaceState *interface = &g_network_model.interfaces[index];
+
+  unsigned int offset = sizeof(RDMHeader);
+  memcpy(g_rdm_buffer + offset, param_data, INTERFACE_ID_SIZE);
+  offset += INTERFACE_ID_SIZE;
+  g_rdm_buffer[offset++] = UInt32Byte0(interface->current_ip);
```
peternewman: Again

firmware/src/network_model.c
```diff
@@ -0,0 +1,837 @@
...
+  g_rdm_buffer[offset++] = UInt32Byte1(interface->current_ip);
+  g_rdm_buffer[offset++] = UInt32Byte2(interface->current_ip);
+  g_rdm_buffer[offset++] = UInt32Byte3(interface->current_ip);
+  g_rdm_buffer[offset++] = interface->current_netmask;
+  if (index == IPSEC_INTERFACE_ID) {
+    g_rdm_buffer[offset++] = DHCP_MODE_UNKNOWN;
```
peternewman: Shouldn't this be NONE if it doesn't support DHCP

nomis52: You'd think so but the standard says otherwise.

firmware/src/network_model.c
```diff
@@ -0,0 +1,837 @@
...
+
+int NetworkModel_GetHostname(const RDMHeader *header,
+                             UNUSED const uint8_t *param_data) {
+  return RDMResponder_GenericReturnString(header, g_network_model.hostname,
+                                          DNS_HOST_NAME_SIZE);
+}
+
+int NetworkModel_SetHostName(const RDMHeader *header,
+                             const uint8_t *param_data) {
+  if (header->param_data_length == 0) {
```
peternewman: Or > DNS_HOST_NAME_SIZE

firmware/src/network_model.c
```diff
@@ -0,0 +1,837 @@
...
+
+int NetworkModel_GetDomainName(const RDMHeader *header,
+                               UNUSED const uint8_t *param_data) {
+  return RDMResponder_GenericReturnString(header, g_network_model.domain_name,
+                                          DNS_DOMAIN_NAME_SIZE);
+}
+
+int NetworkModel_SetDomainName(const RDMHeader *header,
+                               const uint8_t *param_data) {
+  RDMUtil_StringCopy(g_network_model.domain_name, DNS_DOMAIN_NAME_SIZE,
```
peternewman: > DNS_DOMAIN_NAME_SIZE

firmware/src/network_model.c
nomis52: Done

firmware/src/network_model.c
nomis52: sorry??

firmware/src/utils.h
```diff
@@ -65,6 +65,55 @@ static inline uint16_t JoinShort(uint8_t upper, uint8_t lower) {
   return (upper << 8) + lower;
 }
 
+/**
+ * @brief Return the MSB from a 32bit value.
+ * @param s The 32bit value to split.
+ * @returns The first 8 bits of the uint32_t.
+ */
+static inline uint8_t UInt32Byte0(uint32_t s) {
+  return s >> 24;
```
peternewman: Shouldn't these all really have a mask too?

firmware/src/network_model.c
nomis52: changed to INTERFACE_ID_SIZE

firmware/src/utils.h
nomis52: The fact that we return uint8_t handles that

firmware/src/network_model.c
peternewman: There's a > at the start; GitHub has eaten it and turned it into an indent.

nomis52: Ah, that won't ever fail because the max size is equal to the max param data length, but I added it anyway

firmware/src/utils.h
peternewman: Ah okay

firmware/src/network_model.c
peternewman: Where? Section 4.4 says "If the interface does not support DHCP then GET requests shall always return a value of 0 (disabled)". I assume you're referring to 4.6's "Some devices may not be able to determine if the current address was obtained via DHCP. If the device does not support this feature, the value of the DHCP Mode field shall be DHCP_MODE_UNKNOWN." If it's a semicolon rather than a full stop after DHCP, which I suspect it should be, then it makes more sense and still says INACTIVE.

firmware/src/network_model.c
peternewman: Thanks

firmware/src/network_model.c
```diff
@@ -0,0 +1,829 @@
...
+  return RDMResponder_BuildSetAck(header);
+}
+
+int NetworkModel_GetHostname(const RDMHeader *header,
+                             UNUSED const uint8_t *param_data) {
+  return RDMResponder_GenericReturnString(header, g_network_model.hostname,
+                                          DNS_HOST_NAME_SIZE);
+}
+
+int NetworkModel_SetHostName(const RDMHeader *header,
```
peternewman: SetHostname

firmware/src/network_model.c
```diff
@@ -0,0 +1,829 @@
...
+  uint32_t ip = JoinUInt32(param_data[1], param_data[2],
+                           param_data[3], param_data[4]);
+
+  g_network_model.nameservers[index] = ip;
+  return RDMResponder_BuildSetAck(header);
+}
+
+int NetworkModel_GetHostname(const RDMHeader *header,
+                             UNUSED const uint8_t *param_data) {
+  return RDMResponder_GenericReturnString(header, g_network_model.hostname,
```
peternewman: We should really handle the NR_HARDWARE_FAULT if hostname > 63 here.

firmware/src/network_model.c
```diff
@@ -0,0 +1,829 @@
...
+
+  RDMUtil_StringCopy(g_network_model.hostname, DNS_HOST_NAME_SIZE,
+                     (const char*) param_data,
+                     header->param_data_length);
+  return RDMResponder_BuildSetAck(header);
+}
+
+int NetworkModel_GetDomainName(const RDMHeader *header,
+                               UNUSED const uint8_t *param_data) {
+  return RDMResponder_GenericReturnString(header, g_network_model.domain_name,
```
peternewman: We should really handle the NR_HARDWARE_FAULT if hostname > 231 here.

firmware/src/network_model.c
nomis52: The new text is:

"Some devices may not be able to determine if the current address was obtained via DHCP. If the device does not support this feature, the value of the DHCP Status field shall be DHCP_STATUS_UNKNOWN."

I can see how this can be read both ways. It's not clear if 'this feature' means DHCP or the ability to determine if the address was assigned via DHCP. 

I'll add this to my notes.

firmware/src/network_model.c
nomis52: Done

firmware/src/network_model.c
nomis52: The name can't exceed 231 because that's all the space we have to store it. The HARDWARE_FAULT occurs when the name is set by some other mechanism (CLI or the like).

firmware/src/network_model.c
nomis52: Changed to DHCP_STATUS_INACTIVE

firmware/src/network_model.c
peternewman: If we add DHCP support in future, we could get a domain name via that which breaks this.

peternewman:
LGTM

PR #91, opened 2015-07-23
Move the get param length checking into the common function.

nomis52:


nomis52:
This along with #91 reduces the code size by 3%.

peternewman:
Fix Travis.

firmware/src/network_model.c
```diff
@@ -333,14 +290,11 @@ int NetworkModel_SetDHCPMode(const RDMHeader *header,
 
 int NetworkModel_GetZeroconfMode(const RDMHeader *header,
                                  const uint8_t *param_data) {
-  unsigned int index = 0;
```
peternewman: Can you have it more like this to simplify further and remove the duplicate Nack code too?

firmware/src/network_model.c
```diff
@@ -377,59 +329,50 @@ int NetworkModel_SetZeroconfMode(const RDMHeader *header,
...
-  unsigned int offset = sizeof(RDMHeader);
-  memcpy(g_rdm_buffer + offset, param_data, INTERFACE_ID_SIZE);
-  offset += INTERFACE_ID_SIZE;
-  g_rdm_buffer[offset++] = UInt32Byte0(interface->current_ip);
-  g_rdm_buffer[offset++] = UInt32Byte1(interface->current_ip);
-  g_rdm_buffer[offset++] = UInt32Byte2(interface->current_ip);
-  g_rdm_buffer[offset++] = UInt32Byte3(interface->current_ip);
-  g_rdm_buffer[offset++] = interface->current_netmask;
+  uint8_t *ptr = g_rdm_buffer + sizeof(RDMHeader);
+  memcpy(ptr, param_data, INTERFACE_ID_SIZE);
```
peternewman: Is there any reason this one is memcpying and the others aren't?

firmware/src/network_model.c
```diff
@@ -439,16 +382,13 @@ int NetworkModel_SetStaticAddress(const RDMHeader *header,
...
-  int index = LookupIndex(id);
+  const int index = LookupIndex(ExtractUInt32(param_data));
   if (index < 0) {
     return RDMResponder_BuildNack(header, NR_DATA_OUT_OF_RANGE);
   }
 
-  uint32_t ip = JoinUInt32(param_data[4], param_data[5],
-                           param_data[6], param_data[7]);
-  uint8_t netmask = param_data[8];
+  const uint32_t ip = ExtractUInt32(&param_data[4]);
```
peternewman: INTERFACE_ID_SIZE

firmware/src/network_model.c
```diff
@@ -439,16 +382,13 @@ int NetworkModel_SetStaticAddress(const RDMHeader *header,
...
+  const uint8_t netmask = param_data[8];
```
peternewman: sizeof(uint32_t)

firmware/src/network_model.c
nomis52: I'm not following.

The return-if makes things tricky. I don't like introducing macros.

firmware/src/network_model.c
```diff
@@ -565,10 +487,8 @@ int NetworkModel_SetDefaultRoute(const RDMHeader *header,
     return RDMResponder_BuildNack(header, NR_FORMAT_ERROR);
   }
 
-  uint32_t interface_id = JoinUInt32(param_data[0], param_data[1],
-                                     param_data[2], param_data[3]);
-  uint32_t ip = JoinUInt32(param_data[4], param_data[5], param_data[6],
-                           param_data[7]);
+  const uint32_t interface_id = ExtractUInt32(param_data);
+  const uint32_t ip = ExtractUInt32(&param_data[4]);
```
peternewman: Again

firmware/src/network_model.c
nomis52: I don't think that improves readability. It's easy for me to check this is doing the right thing by looking at the offsets, rather than following a const.

firmware/src/network_model.c
nomis52: Same comment here.

firmware/src/network_model.c
```diff
@@ -616,15 +526,12 @@ int NetworkModel_SetNameServer(const RDMHeader *header,
...
+  const uint8_t index = param_data[0];
   if (index >= NUMBER_OF_NAMESERVERS) {
     return RDMResponder_BuildNack(header, NR_DATA_OUT_OF_RANGE);
   }
 
-  uint32_t ip = JoinUInt32(param_data[1], param_data[2],
-                           param_data[3], param_data[4]);
-
-  g_network_model.nameservers[index] = ip;
+  g_network_model.nameservers[index] = ExtractUInt32(&param_data[1]);
```
peternewman: sizeof(uint8_t)

firmware/src/network_model.c
nomis52: We reuse the interface id from the request so that we don't have to keep another variable around.  This reduces the amount of code on line 332

firmware/src/network_model.c
```diff
@@ -765,44 +672,47 @@ const ModelEntry NETWORK_MODEL_ENTRY = {
...
     (PIDCommandHandler) NULL},
-  {PID_IDENTIFY_DEVICE, RDMResponder_GetIdentifyDevice,
+  {PID_IDENTIFY_DEVICE, RDMResponder_GetIdentifyDevice, 0,
     RDMResponder_SetIdentifyDevice},
-
-  {PID_LIST_INTERFACES, NetworkModel_GetListInterfaces,
+  {PID_LIST_INTERFACES, NetworkModel_GetListInterfaces, 0,
     (PIDCommandHandler) NULL},
-  {PID_INTERFACE_LABEL, NetworkModel_GetInterfaceLabel,
+  {PID_INTERFACE_LABEL, NetworkModel_GetInterfaceLabel, 4,
```
peternewman: INTERFACE_ID_SIZE

nomis52: I have other plans for this so I'm going to leave it as is for now.

firmware/src/rdm_handler.c
```diff
@@ -132,20 +125,17 @@ static int GetModelList(const RDMHeader *header) {
...
-      g_rdm_buffer[offset++] = ShortLSB(g_models[i].model_id);
-      size+= 2;
+      ptr = PushUInt16(ptr, g_models[i].model_id);
     }
-    i++;
   }
 
   RDMResponder_BuildHeader(header, ACK, GET_COMMAND_RESPONSE,
-                           PID_DEVICE_MODEL_LIST, size);
+                           PID_DEVICE_MODEL_LIST,
```
peternewman: Fetch from param data?

nomis52: This actually gets replaced in the next PR to address this very problem :)

firmware/src/rdm_responder.c
```diff
@@ -352,11 +333,12 @@ int RDMResponder_SetMute(const RDMHeader *header) {
   g_responder.is_muted = true;
 
   ReturnUnlessUnicast(header);
+
+  uint8_t *ptr = g_rdm_buffer + sizeof(RDMHeader);
+  ptr = PushUInt16(ptr, 0);  // set control fiels to 0
```
peternewman: SPaG fields

firmware/src/rdm_responder.c
```diff
@@ -368,25 +350,22 @@ int RDMResponder_SetUnMute(const RDMHeader *header) {
   g_responder.is_muted = false;
 
   ReturnUnlessUnicast(header);
+
+  uint8_t *ptr = g_rdm_buffer + sizeof(RDMHeader);
+  ptr = PushUInt16(ptr, 0);  // set control fiels to 0
```
peternewman: Again

firmware/src/rdm_responder.c
```diff
@@ -722,53 +625,26 @@ int RDMResponder_GetSensorDefinition(const RDMHeader *header,
...
+  *ptr++ = sensor_ptr->type;
+  *ptr++ = sensor_ptr->unit;
+  *ptr++ = sensor_ptr->prefix;
+  ptr = PushUInt16(ptr, sensor_ptr->range_minimum_value);
+  ptr = PushUInt16(ptr, sensor_ptr->range_maximum_value);
+  ptr = PushUInt16(ptr, sensor_ptr->normal_minimum_value);
+  ptr = PushUInt16(ptr, sensor_ptr->normal_maximum_value);
+
+  ptr += RDMUtil_StringCopy((char*) ptr, RDM_DEFAULT_STRING_SIZE,
+                             sensor_ptr->description, RDM_DEFAULT_STRING_SIZE);
```
peternewman: De-dent one.

firmware/src/rdm_responder.c
```diff
@@ -808,16 +685,18 @@ int RDMResponder_SetSensorValue(const RDMHeader *header,
...
-    BuildSensorValueResponse(sensor_index, &g_responder.sensors[sensor_index]);
+    ptr = BuildSensorValueResponse(ptr, sensor_index,
+                                   &g_responder.sensors[sensor_index]);
   }
 
-  RDMResponder_BuildHeader(header, ACK, SET_COMMAND_RESPONSE,
+  RDMResponder_BuildHeader(header, ACK, GET_COMMAND_RESPONSE,
                            ntohs(header->param_id),
-                           SENSOR_VALUE_PARAM_DATA_LENGTH);
+                           (ptr - g_rdm_buffer) - sizeof(RDMHeader));
```
peternewman: Reviewed to here.

peternewman:
Add a RDMResponder_BuildGetAck which does param id, checksum etc, like RDMResponder_BuildSetAck.

firmware/src/rdm_responder.c
nomis52: Fixed

firmware/src/rdm_responder.c
nomis52: Done

firmware/src/rdm_util.h
```diff
@@ -68,15 +68,14 @@ bool RDMUtil_RequiresAction(const uint8_t our_uid[UID_LENGTH],
 /*
  * @brief Check if we should respond to an RDM request.
  * @param responder The responder information
- * @param uid The UID to check.
+ * @param uid The destination UID to check.
  * @returns true if the UIDs match, false otherwise.
+ * @pre This assumes that RDMUtil_RequiresAction already returned true.
  *
- * A response is only required if the UID exactly matches our UID.
+ * A response is only required if the last for bytes of the UID are not
```
peternewman: SPaG four

firmware/src/utils.h
```diff
@@ -107,16 +107,47 @@ static inline uint8_t UInt32Byte3(uint32_t s) {
...
+ * @param value The value to push.
+ * @returns A pointer to the next byte after the last one that was copied.
+ */
+static inline uint8_t* PushUInt16(uint8_t *ptr, uint16_t value) {
+  *ptr++ = (value >> 8);
+  *ptr++ = value & 0xff;
+  return ptr;
+}
+
+/*
```
peternewman: **

firmware/src/network_model.c
peternewman: Do you mean you don't like:
unsigned int index = 0;
int r = CheckInterfaceRequest(header, param_data, &index);
if (r >= 0) {
  return r;
}

firmware/src/rdm_util.h
nomis52: Fixed

firmware/src/utils.h
nomis52: Fixed

firmware/src/network_model.c
peternewman: In which case how come we're not doing that in all the other cases?

firmware/src/network_model.c
nomis52: Yeah, that's 5 lines instead of 4. It also relies on overloading the convention of what a negative response size represents.

firmware/src/network_model.c
peternewman: But it explains why it's jumping by 4. How about sizeof(uint32_t) instead then?

firmware/src/network_model.c
nomis52: What other cases? I think I do it everywhere.

firmware/src/network_model.c
peternewman: Okay

firmware/src/rdm_handler.c
peternewman: Cool

peternewman:
All reviewed now, just a few outstanding comments.

firmware/src/network_model.c
nomis52: But then things get messy:

sizeof(uint8_t) + sizeof(uint32_t)

Adding a level of indirection makes it harder to figure out what's going on

firmware/src/network_model.c
peternewman: Okay, fair enough.

firmware/src/network_model.c
peternewman: I can't find any now, so either I imagined it, or you've fixed it.

firmware/src/network_model.c
peternewman: Hmm, I'm just not really a fan of a magic number in the code. How about a comment as a compromise?

nomis52: What sort of comment do you suggest ? I think it's going to get quite messy.

Ideally we'd do this with packed structs, but in this case we don't want to pay the penalty of copying memory.

peternewman:
LGTM, I think just one possible suggestion.

PR #97, opened 2015-07-26
Minor cleanup after Eric's feedback.

nomis52:


firmware/src/rdm_responder.c
```diff
@@ -412,7 +412,12 @@ int RDMResponder_GetDeviceInfo(const RDMHeader *header,
   ptr = PushUInt32(ptr, g_responder.def->software_version);
   ptr = PushUInt16(ptr, personality ? personality->dmx_footprint : 0);
   *ptr++ = g_responder.current_personality;
-  *ptr++ = g_responder.def->personality_count;
+
+  if (g_responder.def->personalities) {
+    *ptr++ = g_responder.def->personality_count;
+  } else {
+    *ptr++ = 1;
```
peternewman: Really? If a device has no footprint, how can it have a personality?

peternewman:
Generally LGTM

firmware/src/rdm_responder.c
nomis52: It's open to interpretation. From talking to people it seems 1 / 1 is the unofficial interpretation. I've added it to the fix it list for E1.20.

peternewman: Okay

PR #100, opened 2015-07-29
More cleanup based on MISRA style guide.

nomis52:
https://en.wikipedia.org/wiki/Motor_Industry_Software_Reliability_Association

### DIFF
--- a/firmware/src/network_model.c
+++ b/firmware/src/network_model.c
@@ -53,7 +53,7 @@ static const char SOFTWARE_LABEL[] = "Alpha";
 static const char DEFAULT_DEVICE_LABEL[] = "Ja Rule";
 static const char DEFAULT_HOSTNAME[] = "responder";
 static const char DEFAULT_DOMAINNAME[] = "local";
-static const uint32_t DHCP_FAILURE_RATIO = 3;  // Fail 1 / n DHCP requests.
+static const uint32_t DHCP_FAILURE_RATIO = 3u;  // Fail 1 / n DHCP requests.
 
 static const ResponderDefinition RESPONDER_DEFINITION;
 
@@ -169,12 +169,12 @@ static uint32_t GetDHCPAddress(bool can_fail) {
 static void UseZeroconfOrUnassign(InterfaceState *interface) {
   if (interface->current_zeroconf_mode) {
     interface->current_ip = 0xa9fe0000 + (Random_PseudoGet() % 0xfeff);
-    interface->current_netmask = 16;
+    interface->current_netmask = 16u;
     interface->config_source = CONFIG_SOURCE_ZEROCONF;
   } else {
     interface->config_source = CONFIG_SOURCE_NONE;
     interface->current_ip = IPV4_UNCONFIGURED;
-    interface->current_netmask = 0;
+    interface->current_netmask = 0u;
   }
 }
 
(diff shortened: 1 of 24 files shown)


firmware/src/network_model.c
```diff
@@ -260,8 +260,9 @@ int NetworkModel_GetDHCPMode(const RDMHeader *header,
   unsigned int offset = sizeof(RDMHeader);
   memcpy(g_rdm_buffer + offset, param_data, INTERFACE_ID_SIZE);
   offset += INTERFACE_ID_SIZE;
-  g_rdm_buffer[offset++] =
+  g_rdm_buffer[offset] =
```
peternewman: This seems a lot clearer.

firmware/src/rdm_handler.c
```diff
@@ -31,7 +31,7 @@
 #include "utils.h"
 #include "system_pipeline.h"
 
-#define MAX_RDM_MODELS 6
+enum { MAX_RDM_MODELS = 4};
```
peternewman: Not 6 still?

firmware/src/transceiver.c
```diff
@@ -231,17 +231,6 @@ static inline void RebaseTimer(uint16_t last_event) {
                            PLIB_TMR_Counter16BitGet(TMR_ID_3) - last_event);
 }
 
-/*
```
peternewman: No longer required?

firmware/src/transceiver.c
```diff
@@ -322,8 +312,9 @@ void UART_FlushRX() {
 bool UART_RXBytes() {
   while (PLIB_USART_ReceiverDataIsAvailable(g_hw_settings.usart) &&
          g_transceiver.data_index != BUFFER_SIZE) {
-    g_transceiver.active->data[g_transceiver.data_index++] =
+    g_transceiver.active->data[g_transceiver.data_index] =
       PLIB_USART_ReceiverByteReceive(g_hw_settings.usart);
```
peternewman: Indent four from g_tran

firmware/src/transceiver.c
peternewman: Do they not like inlines?

peternewman:
I wonder if we can lint this <digit>u bit somehow?

firmware/src/network_model.c
nomis52: Yup

firmware/src/rdm_handler.c
nomis52: I've bumped it up since I'm added more models.

firmware/src/transceiver.c
nomis52: It was more of a case that this was only used from one location.

firmware/src/transceiver.c
nomis52: Done

nomis52:
I've asked coverity if we can enable MISRA checking on this repo. I think they only offer as it a commercial tool but it's worth a shot.

firmware/src/rdm_handler.c
peternewman: That's what I'm saying, this commit pushes it back down to 4.

firmware/src/transceiver.c
peternewman: Understood.

firmware/src/rdm_handler.c
nomis52: Oh, good point. I'll increase it in the next change again.

peternewman:
Nice, that would be cool if they did it.

peternewman:
LGTM

PR #110, opened 2015-07-31
Add the bare bones proxy model.

nomis52:
Discovery is working. The next step is to ACK TIMER responses from the child
devices and use queued messages to get the actual RDM response.


firmware/src/rdm_responder.c
```diff
@@ -439,8 +450,7 @@ int RDMResponder_SetMute(const RDMHeader *header) {
   ReturnUnlessUnicast(header);
 
   uint8_t *ptr = g_rdm_buffer + sizeof(RDMHeader);
-  ptr = PushUInt16(ptr,
```
peternewman: Still lots of duplication in these two methods.

peternewman:
LGTM

PR #116, opened 2015-08-02
Use a method to switch responder rather than changing the pointer dir…

nomis52:
…ectly.

### DIFF
--- a/firmware/src/rdm_responder.c
+++ b/firmware/src/rdm_responder.c
@@ -184,6 +184,14 @@ void RDMResponder_Tasks() {
   }
 }
 
+void RDMResponder_SwitchResponder(RDMResponder *responder) {
+  g_responder = responder;
+}
+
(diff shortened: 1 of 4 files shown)


firmware/src/rdm_responder.c
```diff
@@ -184,6 +184,14 @@ void RDMResponder_Tasks() {
   }
 }
 
+void RDMResponder_SwitchResponder(RDMResponder *responder) {
+  g_responder = responder;
+}
+
+void RDMResponder_RestoreResponder() {
+  g_responder = &root_responder;
```
peternewman: Where does root_responder get set?

peternewman:
Generally LGTM

firmware/src/rdm_responder.c
nomis52: It's static, see the top of the file

PR #120, opened 2015-08-04
Merge github.com:nomis52/ja-rule into licence

nomis52:


tests/scripts/enforce_licence.py
```diff
@@ -172,10 +172,10 @@ def GetDirectoryLicences(root_dir):
     for sub_dir in subdirs:
       full_path = os.path.join(dir_name, sub_dir)
       relative_path = os.path.relpath(full_path, root_dir)
-      ok = True
+      include_directory = True
       for ignored_directory in IGNORED_DIRECTORIES:
-        ok &= relative_path.startswith(ignored_directory)
-      if ok:
+        include_directory &= relative_path.startswith(ignored_directory)
```
peternewman: Shouldn't this be !?

peternewman:
Merge this into OLA too when it goes in.

tests/scripts/enforce_licence.py
nomis52: No, that's correct as is.

peternewman: Okay, I'm confused, looking at the code it seems in pseudo code, for each subdir for each ignored dir, check if it matches all the ignored dirs, if it does, add it to the array.

nomis52: Ah, you're correct. PR opened

peternewman: :)

PR #126, opened 2015-08-08
Convert dos line endings to unix.

nomis52:
Also add a simple blinking LED so I can tell when the bootloader is running
while I'm testing.


peternewman:
LGTM, what I can see of the LED bit anyway.

PR #128, opened 2015-08-12
DFU Tools

nomis52:


tools/Makefile
```diff
@@ -0,0 +1,14 @@
+CC=gcc
```
peternewman: Shouldn't this use some configure magic?

tools/constants.h
```diff
@@ -0,0 +1,25 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Library General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Copyright (C) 2015 Simon Newton.
```
peternewman: Do we not add the filename in the header comment here?

tools/Makefile
nomis52: I'm going to move configure to the top level , probably as the last change in this batch.

autotools will then be responsible for:
 - building the dfu tools
 - running the tests
 - generating the documentation.

tools/dfu.c
```diff
@@ -0,0 +1,193 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * CRC polynomial 0xedb88320 – Contributed unknowingly by Gary S. Brown.
+ * "Copyright (C) 1986 Gary S. Brown. You may use this program, or code or
+ * tables extracted from it, as desired without restriction."
+ *
+ * The updcrc macro (referred to here as CalculateCRC) is derived from an
+ * article Copyright 1986 by Stephen Satchell.
```
peternewman: SPaG copyright

tools/dfu.c
```diff
@@ -0,0 +1,193 @@
...
+ *
+ * Remainder portions of the code are Copyright (C) 2015 Simon Newton.
```
peternewman: Again

tools/dfu.c
```diff
@@ -0,0 +1,193 @@
...
+    .vendor_id = htons(options->vendor_id),
+    .product_id = htons(options->product_id),
+    .device = 0xffff,
+  };
+
+  // Write the suffix out backwards.
+  const uint8_t *suffix_ptr = (const uint8_t*) &suffix;
+  for (int i = sizeof(suffix) - 1; i >= 0; i--) {
+    crc = CalculateCRC(crc, suffix_ptr + i, 1);
+    ssize_t r = write(fd, suffix_ptr + i, 1);
```
peternewman: Any reason this can't use Write?

tools/dfu.h
```diff
@@ -0,0 +1,52 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Library General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Copyright (C) 2015 Simon Newton.
```
peternewman: Again, add the filename?

tools/flash_hex.sh
```diff
@@ -0,0 +1,48 @@
...
+$hex2dfu $hex_file;
+
+temp=$($DFU_UTIL --help > /dev/null 2>&1)
+
+if [ $? -ne 0 ]; then
+  echo "Can't execute $DFU_UTIL";
+  exit;
+fi
+
+$hex2dfu $hex_file
```
peternewman: This is a duplicate of line 26.

tools/hex2dfu.c
```diff
@@ -0,0 +1,395 @@
...
+
+/**
+ * @brief Convert 4 characters to a uint16_t.
+ */
+bool HexToUInt16(const char *str, uint16_t *output) {
+  uint8_t upper, lower;
+  if (!HexToUInt8(str, &upper) || !HexToUInt8(str + 2, &lower)) {
+    return false;
+  }
+  *output = (upper << 8) + lower;
```
peternewman: JoinUInt8

tools/uid2dfu.c
```diff
@@ -0,0 +1,161 @@
...
+  options->device_id = 0;
+  options->help = false;
+
+  bool got_device = false;
+  bool got_manufacturer = false;
+
+  static struct option long_options[] = {
+      {"device", required_argument, 0, 'd'},
+      {"help", no_argument, 0, 'h'},
+      {"manufacturer", required_argument, 0, ','},
```
peternewman: Shouldn't the comma be an m?

tools/uid2dfu.c
```diff
@@ -0,0 +1,161 @@
...
+  bool help;
+} Options;
+
+static const char DEFAULT_FILE[] = "uid.dfu";
+
+void DisplayHelpAndExit(const char *arg0, int exit_code) {
+  printf("Usage: %s [options] -m <manufacturer-id> -d <device-id>\n", arg0);
+  printf("  -d, --device <id>  The device ID\n");
+  printf("  -h, --help   Show the help message\n");
+  printf("  -m, --manufacturer <id>  The manufacturer ID\n");
```
peternewman: Any reason the manufacturer ID shouldn't default to us?

tools/Makefile
peternewman: Cool

peternewman:
Generally LGTM

tools/constants.h
nomis52: Added

tools/dfu.c
nomis52: Done

tools/dfu.c
nomis52: Done

tools/dfu.c
nomis52: Changed

tools/dfu.h
nomis52: Added

tools/flash_hex.sh
nomis52: Good point, removed the first one.

tools/hex2dfu.c
nomis52: There is no such function here and I don't want to depend on the OLA libs.

tools/uid2dfu.c
nomis52: Fixed.

tools/uid2dfu.c
nomis52: I'm a bit cautious of people (manufacturers) running this script and accidentally using our ID.

tools/hex2dfu.c
peternewman: Okay, I thought we had similar in other bits of the Ja Rule code that we could pull in.

nomis52: On the embedded side yes, but this is host side software.

tools/uid2dfu.c
peternewman: Okay. What are we doing about reserving UIDs globally anyway?

peternewman: How about as a compromise, when you add configure support, adding an option that configures a default manufacturer ID, but defaults to none/or one of the test ones.

nomis52: We're not putting our ID in hardware devices. Manufacturers have to use their own manufacturer ID.

nomis52: That's a bit scary though if people start distributing the binaries. Ideally one day we'll include these in .debs / macports.

tools/flash_hex.sh
```diff
@@ -0,0 +1,46 @@
...
+tool_dir=$(dirname $0)
+hex2dfu="${tool_dir}/hex2dfu"
+
+if [ ! -x $hex2dfu ]; then
+  echo "Can't execute $hex2dfu, try running make first";
+  exit;
+fi
+
+$DFU_UTIL --help > /dev/null 2>&1
+
```
peternewman: For consistency, can we remove the space here please.

peternewman:
LGTm

tools/flash_hex.sh
nomis52: Which space ?

peternewman: Sorry, I meant blank line

nomis52: Done

tools/hex2dfu.c
peternewman: Ah, I assumed you could just import it still, or do you not want to?

tools/uid2dfu.c
peternewman: Ah, I hadn't really considered that, I assumed we'd just be using our ID. So what happens if some random person makes a PCB from the design, or uses the dev kit you had, do they have to get a manufacturer ID from PLASA?

tools/hex2dfu.c
nomis52: Yeah I don't want to mix the two.

tools/uid2dfu.c
nomis52: Correct. If you're making and selling hardware you need your own.

We could look at sub-allocating blocks of IDs to hobbyists but I'd like to put that off as long as we can due to the admin burden.

tools/hex2dfu.c
peternewman: Okay

peternewman:
LGTM

tools/uid2dfu.c
peternewman: Are we assuming no-one will ever make production hardware/low volume/one off specials using this then?

Playing devil's advocate, a few interesting questions:
If I buy a unit, will it come with a UID printed on?
What happens if I run the proxy demo code on it, or make a Lab Pack style unit, where do I get the other UIDs from?
Keeping model IDs unique within the manufacturer code (and someone running the code you've written, with set model IDs, against another manufacturer's UID)
General code quality, if the software isn't written by the hardware manufacturer

nomis52: > Are we assuming no-one will ever make production hardware/low volume/one off specials using this then?

It comes down to 'if you're selling units, you need your own ID'. I believe PLASA has now set aside some UIDs for prototyping use so I'd prefer people to use those rather than our id-space. 

> If I buy a unit, will it come with a UID printed on?

It's up to the manufacturer. It does has to be assigned a UID though if they want to use RDM. It would make sense to print the UID on the board / case.

> What happens if I run the proxy demo code on it, or make a Lab Pack style unit, where do I get the other UIDs from?

Manufacturers should leave the last 4 bits of the UID set to 0 to allow for this. We should document this somewhere.

PR #133, opened 2015-08-23
Add the common files.

nomis52:
This includes the API to fetch the UID from flash and the API to check / get
the boot option.

### DIFF
--- a/common/dfu_constants.h
+++ b/common/dfu_constants.h
@@ -1,0 +1,112 @@
+/*
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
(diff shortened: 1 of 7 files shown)


common/dfu_constants.h
```diff
@@ -0,0 +1,112 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief DFU message types
+ */
+enum {
+  DFU_DETATCH = 0,
```
peternewman: Didn't we have this already, or have you moved it?

peternewman:
Generally LGTM

common/dfu_constants.h
nomis52: We have Bootloader/firmware/src/dfu_constants.h but that's going to get removed in an upcoming PR.

peternewman: Ah

PR #136, opened 2015-08-25
Add the reset module and update the licence script

nomis52:


common/reset.c
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * reset.c
+ * Copyright (C) 2015 Simon Newton
+ */
+
+#include "reset.h"
+#include <peripheral/reset.h>
+
+void Reset_SoftReset() {
+  SoftReset();
```
peternewman: I take it this is just to abstract the underlying method?

peternewman:
LGTM

common/reset.c
nomis52: Yup, it's so we can mock it out for tests. Same with the launcher code - jumping to a random memory address is not useful :)

PR #137, opened 2015-08-25
Cleanup tests before we start adding the bootloader

nomis52:


firmware/src/usb_transport.c
```diff
@@ -298,7 +299,7 @@ bool USBTransport_SendResponse(uint8_t token, Command command, uint8_t rc,
       memcpy(transmitDataBuffer + offset + 8, data[i].base,
              PAYLOAD_SIZE - offset);
       offset = PAYLOAD_SIZE;
-      transmitDataBuffer[6] |= TRANSPORT_MSG_TRUNCATED;
+      transmitDataBuffer[7] |= TRANSPORT_MSG_TRUNCATED;
```
peternewman: Is it worth making the 7 a constant?

peternewman:
LGTM

firmware/src/usb_transport.c
nomis52: I don't think so.

PR #148, opened 2015-08-30
Number8

nomis52:


firmware/ja-rule.X/nbproject/configurations.xml
```diff
@@ -287,46 +287,178 @@
         <makeCustomizationNormalizeHexFile>false</makeCustomizationNormalizeHexFile>
       </makeCustomizationType>
       <item path="../../linker/app_mx795F512L.ld" ex="true" overriding="false">
+        <C32>
```
peternewman: I assume this is autogenerated by Microchip?

peternewman:
LGTM

firmware/ja-rule.X/nbproject/configurations.xml
nomis52: Yup, all of this is managed by the IDE.

PR #155, opened 2015-09-03
Move hardware peripheral settings to a separate file.

nomis52:


peternewman:
Fix Travis

tests/Doxyfile
```diff
@@ -656,6 +656,7 @@ WARN_LOGFILE           =
 INPUT                  = ../Bootloader/firmware/src \
                          ../common \
                          ../firmware/src \
+                         ../firmware/src/system_config/ethernet_sk2/system_settings.h \
```
peternewman: Add both system_settings.h?

peternewman:
Generally LGTM

tests/Doxyfile
nomis52: Nope, just one. Otherwise we'll get duplicate comments.

peternewman: How about adding tests/system_config/system_settings.h or at least the number 8 one instead then, rather than the dev one, which is less likely to get updates to Doxygen?

peternewman:
LGTM

tests/Doxyfile
nomis52: I realized last night I'm going to have to move this around a bit, since there are settings shared between the bootloader and the main app.

I think what I'll end up doing is creating a top level board_cfg directory, and then within that we'll have a template directory which people can copy when porting.

peternewman: Sounds good, and avoids duplication of Doxygen, I was going to suggest something in that direction. If we go to Autotools, can we then just include the relevant board config depending on what you pick during configure.

PR #164, opened 2015-09-07
Move board config to it's own directory.

nomis52:

PR #165, opened 2015-09-07
Add some missing linting

peternewman:


peternewman:
I think this will catch issues like https://github.com/OpenLightingProject/ja-rule/pull/164/files#diff-7dbd725f9a103e05d890e8583ca15ae4L21 and https://github.com/OpenLightingProject/ja-rule/pull/164/files#diff-bd20691697e67aa595e9a175355e6f0aL51 so can we get it in before https://github.com/OpenLightingProject/ja-rule/pull/164 to check please.

nomis52:
Merging, then I'll fix the lint issues

PR #166, opened 2015-09-08
Check for generic NOLINTs with Travis

peternewman:


peternewman:
Can we get this in before https://github.com/OpenLightingProject/ja-rule/pull/164

.travis-ci.sh
```diff
@@ -9,6 +9,17 @@ COVERITY_SCAN_BUILD_URL="https://scan.coverity.com/scripts/travisci_build_coveri
 
 if [[ $TASK = 'lint' ]]; then
   # run the lint tool only if it is the requested task
+  # first check we've not got any generic NOLINTs
+  # count the number of generic NOLINTs
+  nolints=$(grep -IR NOLINT * | grep -v "NOLINT(" | wc -l)
+  if [[ $nolints -ne 0 ]]; then
+    # print the output for info
+    echo $(grep -IR NOLINT * | grep -v "NOLINT(")
```
nomis52: Try 

 grep -IR 'NOLINT\s*$' * 

instead

peternewman: Does it have to be at the end of the line to work for cpplint though, I got the feeling you could write some more words after it in the comment and it would still be fine?

peternewman: I guess NOLINT[^\(]* or NOLINT[^\(]*.*$ ought to work, but I'd want to test it first.

nomis52: I haven't ever seen it not at the end of the line.

peternewman: I can't see anything in the cpplint regex saying it has to be on a line ending though:
https://github.com/google/styleguide/blob/gh-pages/cpplint/cpplint.py#L517

It would seem negative lookahead avoids the extra grep line, this seems to work:

peternewman: Needs -P in grep, but that appears fine.

peternewman:
I think this is good to merge now @nomis52

nomis52:
LGTM

PR #172, opened 2015-09-10
Improve licence enforcer coverage

peternewman:


tests/scripts/enforce_licence.py
```diff
@@ -28,24 +28,51 @@
...
   'src/main.c'
 ]
 
 IGNORED_DIRECTORIES = [
-    'Bootloader/firmware/src/system_config',
-    'firmware/src/system_config',
-    'src/system_config',
-    'tests/src/system_config',
-    'tests/boot_src/system_config',
+    'Bootloader/firmware/src/system_config/ethernet_sk2/framework/',
```
nomis52: Change to Bootloader/firmware/src/system_config & firmware/src/system_config

tests/scripts/enforce_licence.py
```diff
@@ -28,24 +28,51 @@
...
+  'firmware/src/system_config/ethernet_sk2/system_init.c',
+  'firmware/src/system_config/ethernet_sk2/system_interrupt.c',
+  'firmware/src/system_config/ethernet_sk2/system_pipeline.h',
+  'firmware/src/system_config/ethernet_sk2/system_tasks.c',
+  'firmware/src/system_config/number8/system_config.h',
+  'firmware/src/system_config/number8/system_definitions.h',
+  'firmware/src/system_config/number8/system_init.c',
+  'firmware/src/system_config/number8/system_interrupt.c',
+  'firmware/src/system_config/number8/system_pipeline.h',
+  'firmware/src/system_config/number8/system_tasks.c',
```
nomis52: Do we need these if we've added the directories below? I hope not.

tests/scripts/enforce_licence.py
peternewman: That's what you had, but there are valid licensed files in there that should be checked. I've just explicitly blacklisted the autogenerated stuff.

nomis52: There shouldn't be any now, those directories are all autogenerated,

peternewman: Ah yes, I was out of step, it's all in boardcfg now isn't it?

nomis52: Yup

peternewman: Fixed

tests/scripts/enforce_licence.py
peternewman: Removed.

nomis52:
Can you move the scripts directory to the top level as part of this as well. I've done everything else.

peternewman:
Moved the file.

peternewman:
Just thinking @nomis52 given we need to somehow handle the path depth for licence header comments in boardconfig, shall I extend the enforcer to specify a folder and depth, and it will then complain if you don't have the right depth for the name within that folder?

nomis52:
It may be easier just to change the file paths to remove the directories :)

peternewman:
You mean within the file headers?

nomis52:
Yeah

change

number8/board_init.h to just board_init.h

peternewman:
I thought you wanted the folder in place for clarity? I don't think it should be too hard to code up if you want to keep it.

nomis52:
It's not worth keeping it it causes trouble.

peternewman:
It was easy enough to fix within the licence enforcer script, so that's now done (and a few bits it caught fixed). Merge if you're happy @nomis52

scripts/enforce_licence.py
```diff
@@ -249,7 +255,20 @@ def CheckLicenceForFile(file_name, licence, lang, diff, fix):
...
-    expected_line = TransformLine(os.path.basename(file_name), lang)
+    relative_path = os.path.relpath(os.path.dirname(file_name), os.getcwd())
+    expected_filename_depth = 0
+    for dir_name, depth in FILENAME_DEPTH_OVERRIDE.iteritems():
+      if relative_path.startswith(dir_name):
+        expected_filename_depth = depth
+    expected_filename_parts = [os.path.basename(file_name)]
+    depth_path = relative_path
+    for i in range(expected_filename_depth):
+      (depth_path, folder) = os.path.split(depth_path)
```
nomis52: No need for parens here

nomis52:
Just a single comment. Merge when fixed.

scripts/enforce_licence.py
peternewman: Done.

PR #180, opened 2015-09-15
Add the missing Makefile

nomis52:


peternewman:
LGTM, untested so far (I copied from Ja-Rule Makefile and that seemed to work).

PR #187, opened 2015-09-18
Update main.md

hezmand:
first update


user_manual/docs/main.md
```diff
@@ -5,6 +5,8 @@ Introduction     {#mainpage}
 
 Explain what it is here.
 
+Ja Rule is a rapper like DMX
```
nomis52: Are you sure it's not DMX512?

PR #192, opened 2015-09-25
Add a comment about min version of OS X for USB logging

nomis52:


doxygen/main.md
```diff
@@ -40,7 +40,7 @@ DMX receiver / RDM responder.
 
 ## Common {#main-features-common}
 
-- \ref usb-logging, with adjustable log levels.
+- \ref usb-logging, with adjustable log levels. Requires OS X 10.7 or greater.
```
peternewman: Can we word this in such a way it doesn't imply you need OS X. We should probably also clarify that Windows won't work without a driver file.

nomis52: Done

doxygen/main.md
```diff
@@ -40,7 +40,8 @@ DMX receiver / RDM responder.
 
 ## Common {#main-features-common}
 
-- \ref usb-logging, with adjustable log levels.
+- \ref usb-logging, with adjustable log levels. Requires Linux or MAC OS X 10.7 or
```
peternewman: SPaG Mac

peternewman: Re-ordering to list Mac or Linux would make it even clearer that the 10.7 requirement doesn't apply to Linux.

PR #193, opened 2015-09-26
Allow hex2dfu to work with Windows line ending hex files, closes #181

### LINKED ISSUE #181: hex2dfu doesn't handle DOS output with carriage returns
I can fix the first error:
tools/hex2dfu firmware/ja-rule.X/dist/number8/production/ja-rule.X.production.hex
Missing \n on line 1
337885 bytes remain in hex file

With:
diff --git a/tools/hex2dfu.c b/tools/hex2dfu.c
index b1aee72..a624b6f 100644
--- a/tools/hex2dfu.c
+++ b/tools/hex2dfu.c
@@ -248,8 +248,8 @@ void ProcessHexFile(int fd, const Options *options) {
       return;
     }

-    if (checksum_data[2] != '\n') {
-      printf("Missing \\n on line %d\n", line);
+    if (checksum_data[2] != '\n' && checksum_data[2] != '\r') {
+      printf("Missing \\n or \\r on line %d\n", line);
       return;
     }

But then it throws:
tools/hex2dfu firmware/ja-rule.X/dist/number8/production/ja-rule.X.production.hex
Invalid start code '
' on line 2
337876 bytes remain in hex file

I don't know how we want to go about fixing this?

peternewman:


nomis52:
LGTM

PR #195, opened 2015-09-26
Add checks to ensure people don't load the wrong f/w onto devices.

nomis52:


Bootloader/firmware/src/bootloader.c
```diff
@@ -590,6 +594,13 @@ void ProcessDownload() {
       }
       g_transfer.total_size = total_size;
 
+      uint16_t model_id = ExtractUInt16(g_data_buffer + 2 * sizeof(uint32_t));
```
peternewman: Brackets please

common/models.h
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * @brief Hardware model IDs.
+ *
+ * These are used to check we're using the correct firmware for the device.
+ */
+typedef enum {
+  MODEL_UNDEFINED = 0,  //!< Undefined, disables checks
+  MODEL_NUMBER1 = 1,  //!< Number 1.
+  MODEL_NUMBER8 = 2,  //!< Number 8.
+  MODEL_ETHERNET_SK2 = 3,  //!< Ethernet Starter Kit II
```
peternewman: Is it worth splitting this off into another block out of the way?

common/models.h
```diff
@@ -0,0 +1,45 @@
...
+#define COMMON_MODELS_H_
+
+/**
+ * @brief Hardware model IDs.
+ *
+ * These are used to check we're using the correct firmware for the device.
+ */
+typedef enum {
+  MODEL_UNDEFINED = 0,  //!< Undefined, disables checks
+  MODEL_NUMBER1 = 1,  //!< Number 1.
```
peternewman: Do we need to differentiate between one and two port variants of this, so it doesn't present two ports when only one is pinned out?

tests/system_config/bootloader_settings.h
```diff
@@ -17,11 +17,12 @@
  * Copyright (C) 2015 Simon Newton
  */
 
-#ifndef BOOTLOADER_FIRMWARE_SRC_SYSTEM_CONFIG_ETHERNET_SK2_BOOTLOADER_SETTINGS_H_
-#define BOOTLOADER_FIRMWARE_SRC_SYSTEM_CONFIG_ETHERNET_SK2_BOOTLOADER_SETTINGS_H_
+#ifndef TESTS_SYSTEM_CONFIG_BOOTLOADER_SETTINGS_H_
```
peternewman: Perhaps cpplint should be running on this?

tests/system_config/bootloader_settings.h
```diff
@@ -54,6 +55,11 @@ const PORTS_BIT_POS SWITCH_PORT_BIT = PORTS_BIT_POS_7;
 const bool SWITCH_ACTIVE_HIGH = false;
 
 /**
+ * @brief The hardware model.
+ */
+#define HARDWARE_MODEL MODEL_ETHERNET_SK2
```
peternewman: Is it worth giving this a different MODEL_MOCK or something ID?

peternewman:
Generally LGTM

Bootloader/firmware/src/bootloader.c
nomis52: Not for this, the order is clear.

common/models.h
nomis52: What do you mean? I'd prefer to have all models in once place so they don't collide.

common/models.h
nomis52: Hmm good question. I'm not sure it would really work out though as it would require someone to flash the bootloader to go from 1 to 2 ports.

tests/system_config/bootloader_settings.h
nomis52: Added

tests/system_config/bootloader_settings.h
nomis52: Then we face the collision problem. This is fine.

common/models.h
peternewman: I meant numerically, allocating a block at the top of the range for dev boards or something.

common/models.h
peternewman: Is the assumption they can add the components for the second port themselves then? How is the unit going to tell OLA how many ports its got available, or will they just have two firmware variants to load?

tests/system_config/bootloader_settings.h
peternewman: I meant defined within the core codebase.

common/models.h
nomis52: We can't tell in software, so it's up to the user to load the right f/w for the device.

common/models.h
nomis52: If we find ourselves adding more we can do that.

tests/system_config/bootloader_settings.h
nomis52: No, I don't want to leak test data into the main codebase.

peternewman:
LGTM

PR #197, opened 2015-09-26
Add CRC checking to the bootloader.

### LINKED ISSUE #157: Checksum the firmware before restarting.
There is a checksum field in the firmware header but right now it's ignored. We should use the same CRC as the DFU spec for simplicity. There is an implementation in tools/dfu.c

nomis52:
Closes #157


peternewman:
LGTM

I thought we had the CRC stuff in already, or was that client side?

nomis52:
We were calculating the CRC on the host side but not checking it in the bootloader.

peternewman:
Ah okay

PR #204, opened 2015-10-03
Add 2 #defines which allow for running without the bootloader

nomis52:
This is used for testing & debugging.


peternewman:
LGTM

PR #208, opened 2015-10-05
New board config options.

### LINKED ISSUE #207: Build the UID from the MAC address
EMAC1SA0, EMAC1SA1, EMAC1SA2 

http://ww1.microchip.com/downloads/en/DeviceDoc/61155C.pdf

nomis52:
This adds options for:
 - using the MAC address to derive the UID
 - disabling flashing the UID using DFU.

It also cleans up a lot of the documentation.
Closes #207


peternewman:
Fix Travis

common/config_options.h
```diff
@@ -0,0 +1,50 @@
...
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ *
+ * common/config_options.h
```
peternewman: It's complaining about the common/ on here.

common/config_options.h
```diff
@@ -0,0 +1,50 @@
...
+  MODEL_NUMBER1 = 1,  //!< Number 1.
+  MODEL_NUMBER8 = 2,  //!< Number 8.
+  MODEL_ETHERNET_SK2 = 3,  //!< Ethernet Starter Kit II
+} JaRuleModel;
+
+/**
+ * @brief Controls how the device's UID is derived.
+ */
+typedef enum {
+  UID_FROM_MAC = 0,  //!< Use the MAC address
```
peternewman: Where is the code for this? Did you forget to push?

nomis52: Yes, give me a minute..

Bootloader/firmware/src/system_config/ethernet_sk2/system_init.c
```diff
@@ -178,7 +178,8 @@ void SYS_Initialize(void* data) {
 
     /* Initialize Middleware */
 
-    /* Copy the UID from flash to the USB descriptor. */
+    /* Copy the UID from to the USB descriptor. */
```
peternewman: From what?

Bootloader/firmware/src/system_config/number1/system_init.c
```diff
@@ -178,7 +178,8 @@ void SYS_Initialize ( void* data )
 
     /* Initialize Middleware */
 
-    /* Copy the UID from flash to the USB descriptor. */
+    /* Copy the UID from to the USB descriptor. */
```
peternewman: Again

Bootloader/firmware/src/system_config/number8/system_init.c
```diff
@@ -177,7 +177,8 @@ void SYS_Initialize(void* data) {
 
     /* Initialize Middleware */
 
-    /* Copy the UID from flash to the USB descriptor. */
+    /* Copy the UID from to the USB descriptor. */
```
peternewman: Again

boardcfg/number1/common_settings.h
```diff
@@ -40,24 +48,47 @@
...
- * when running without a bootloader.
+ * @name RDM Identification
+ * Testing
+ * @{
+ */
+
+/**
+ * @brief The RDM manufacturer ID
+ */
+#define CFG_MANUFACTURER_ID 0x7a70
```
peternewman: I thought this was going to be Stellascapes from what you said before?

common/uid_store.c
```diff
@@ -17,22 +17,72 @@
...
+  // The first 3 bytes of the MAC address is the Microchip OID 00:1e:c0 which
+  // is constant. The bottom 3 bytes contain the unique serial number.
+  //
+  // To support more than one responder per device, we set the lower 4 bits of
+  // the UID to 0 so we have 16 responders per device. These means the complete
+  // UID takes the form:
+  //   MMMM:0AAAAAA0
+  // Where M is the PLASA manufacturer ID and A are the values from the MAC
+  // address.
+  kUIDArray[0] = CFG_MANUFACTURER_ID >> 8;
```
peternewman: SplitUInt16

common/uid_store.c
```diff
@@ -17,22 +17,72 @@
...
+
+void UIDStore_Init() {
+  // The UID is derived from the RDM manufacturer ID & the MAC address.
+  // The first 3 bytes of the MAC address is the Microchip OID 00:1e:c0 which
+  // is constant. The bottom 3 bytes contain the unique serial number.
+  //
+  // To support more than one responder per device, we set the lower 4 bits of
+  // the UID to 0 so we have 16 responders per device. These means the complete
+  // UID takes the form:
+  //   MMMM:0AAAAAA0
```
peternewman: Can you update https://wiki.openlighting.org/index.php/Open_Lighting_Allocations#UIDs please. Given the current allocation, are we better off the first digit of the device ID being a 1, so we don't clash with existing allocations?

common/uid_store.c
```diff
@@ -17,22 +17,72 @@
...
+  // Where M is the PLASA manufacturer ID and A are the values from the MAC
+  // address.
+  kUIDArray[0] = CFG_MANUFACTURER_ID >> 8;
+  kUIDArray[1] = CFG_MANUFACTURER_ID & 0xff;
+  kUIDArray[2] = ShiftRight(PLIB_ETH_StationAddressGet(ETH_ID_0, 4));
```
peternewman: We should really check that the other digits are indeed the Microchip allocation, in case they ever overflow and use a new allocation.

common/uid_store.c
```diff
@@ -17,22 +17,72 @@
...
+extern uint8_t __attribute__((space(prog))) _uid;
+const uint8_t* const kUID = &_uid;
+
+void UIDStore_Init() {}
+
+#elif CFG_UID_SOURCE == UID_TEST_UID
+
+static const uint8_t kDevUID[UID_LENGTH] = {
+  CFG_MANUFACTURER_ID >> 8,
+  CFG_MANUFACTURER_ID & 0xff,
```
peternewman: This is currently the first OLA dummy device: https://wiki.openlighting.org/index.php/Open_Lighting_Allocations#UIDs

Why not bump it up to 7a70:fffffffe or into the developement range as 7a70:fffffe00?

peternewman:
A few comments

Bootloader/firmware/src/system_config/ethernet_sk2/system_init.c
nomis52: Done

Bootloader/firmware/src/system_config/number1/system_init.c
nomis52: Done

Bootloader/firmware/src/system_config/number8/system_init.c
nomis52: Done

boardcfg/number1/common_settings.h
nomis52: Yeah, but we're not going to check that in here. It'll be a fork.

common/uid_store.c
nomis52: Not here, the libs aren't available because this code is in the bootloader.

common/uid_store.c
nomis52: This doesn't grant people the ok to use the 7a70 space. I don't want to make it look like there 7a70:0XXXXXXX range is available.

common/uid_store.c
nomis52: That's up to manufacturers. We don't have a good way of handling a fault at this point.

common/uid_store.c
nomis52: Done

boardcfg/number1/common_settings.h
peternewman: Okay

common/uid_store.c
peternewman: We could move those to common/utils.h though, we've already got other stuff loaded in both projects.

common/uid_store.c
peternewman: I still think we should keep a track on the wiki from a project perspective, so we know what's being used where; likewise changing the first digit to a 1 helps us as a project. Perhaps we should also add a note about usage on that wiki page. If we do get abuse FTDI clone style, it would be better if they don't clash with our existing UID allocations anyway.

boardcfg/number1/common_settings.h
peternewman: Out of interest, are the model numbers going to match in Stellascapes land? I guess that's a question for @mrpackethead ?

common/uid_store.c
peternewman: It shouldn't necessarily matter for other manufacturers, if their forks are coded differently, but for anyone using the ESK to test, it would be nice to check. Are we too low level that we can just stall and turn all the LEDs on statically (and I guess ideally log something to the console)?

common/uid_store.c
peternewman: Can we add this block of 16 to the UID allocations list please.

common/uid_store.c
nomis52: Switched to start from 1.

peternewman: Thanks

PR #213, opened 2015-10-10
Fix coverity, closes #196

### LINKED ISSUE #196: Fix coverity and coveralls
They both seem to be broken.

peternewman:
Closes #196


peternewman:
At least I think this will fix it...

PR #224, opened 2015-10-12
Handle multiple microchip OIDs.

nomis52:
We've realized microchip has 3 OIDs. This handles the 2 OIDs we've seen in
production - we don't want to allocate the third unless we start shipping chips
with that OID.

The change also changes the GET_UID command to return the Ja Rule Model & the
MAC address. We'll use this in the production tool.

### DIFF
--- a/common/uid_store.c
+++ b/common/uid_store.c
@@ -45,15 +45,15 @@ inline uint8_t ShiftLeft(uint8_t b) {
 void UIDStore_Init() {
   // The UID is derived from the RDM manufacturer ID & the MAC address.
   // The first 3 bytes of the MAC address is the Microchip OIDs, which are one
-  // of 00:1e:c0, 00:04:A3 or D8:80:39. The bottom 3 bytes contain the unique
+  // of 00:1E:C0, 00:04:A3 or D8:80:39. The bottom 3 bytes contain the unique
   // serial number.
   //
   // To support more than one responder per device, we set the lower 4 bits of
   // the UID to 0 so we have 16 responders per device. These means the complete
   // UID takes the form:
   //   MMMM:XAAAAAA0
-  // Where M is the PLASA manufacturer ID and A are the values from the MAC
-  // address. X is derived from the OID:
+  // Where M is the PLASA manufacturer ID, X is derived from the OID and A is
+  // the values from the MAC address. X is derived from the OID as follows:
   //
   //   00:1E:C0 -> 1
   //   D8:80:39 -> 2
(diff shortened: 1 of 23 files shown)


common/uid_store.c
```diff
@@ -42,23 +44,49 @@ inline uint8_t ShiftLeft(uint8_t b) {
...
+  // The first 3 bytes of the MAC address is the Microchip OIDs, which are one
+  // of 00:1e:c0, 00:04:A3 or D8:80:39. The bottom 3 bytes contain the unique
+  // serial number.
   //
   // To support more than one responder per device, we set the lower 4 bits of
   // the UID to 0 so we have 16 responders per device. These means the complete
   // UID takes the form:
-  //   MMMM:1AAAAAA0
+  //   MMMM:XAAAAAA0
   // Where M is the PLASA manufacturer ID and A are the values from the MAC
```
peternewman: SPaG the and should become a comma, but perhaps also re-order so it's M, X, A.

common/uid_store.c
```diff
@@ -42,23 +44,49 @@ inline uint8_t ShiftLeft(uint8_t b) {
...
+    kUIDArray[1] = CFG_MANUFACTURER_ID & 0xff;
+    kUIDArray[2] = upper_id +
+                   ShiftRight(PLIB_ETH_StationAddressGet(ETH_ID_0, 4));
+    kUIDArray[3] = ShiftLeft(PLIB_ETH_StationAddressGet(ETH_ID_0, 4)) +
+                   ShiftRight(PLIB_ETH_StationAddressGet(ETH_ID_0, 5));
+    kUIDArray[4] = ShiftLeft(PLIB_ETH_StationAddressGet(ETH_ID_0, 5)) +
+                   ShiftRight(PLIB_ETH_StationAddressGet(ETH_ID_0, 6));
+    kUIDArray[5] = ShiftLeft(PLIB_ETH_StationAddressGet(ETH_ID_0, 6));
+  } else {
+    // If we didn't match the OID, default to the NULL UID to make it obvious
```
peternewman: Great, that solves my other worry.

common/uid_store.c
peternewman: Done.

peternewman:
LGTM

peternewman:
Erm, I thought it seemed odd that there would be a TLA collision. You need to s/OID/OUI/g, which probably is easiest done on a computer checkout @nomis52  ...

nomis52:
Fixed in #225

PR #229, opened 2015-10-12
RDM Responder changes.

### LINKED ISSUE #227: g_responder->is_managed_proxy  isn't reset when changing models
This causes us the fail some of the RDM tests if you do

- set model_id 257
- run tests
- set model_id 258
- run tests

I need to think about where to initialize / reset this.

nomis52:
- Only flash the Identify & Mute leds when we're in responder mode
- Correctly reset the control field flags when we switch models.

Closes #227


peternewman:
LGTM

PR #232, opened 2015-10-14
Add a self test command.

nomis52:
This puts the driver circuit into loopback mode, sends a single byte, then
verifies we received it. That way we can check that the 485 driver is working.

Side note: using separate control lines for TX / RX Enable just paid off!!

### DIFF
--- a/firmware/src/app.c
+++ b/firmware/src/app.c
@@ -156,9 +156,9 @@ void APP_Tasks(void) {
   USBTransport_Tasks();
   Transceiver_Tasks();
   USBConsole_Tasks();
-  RDMResponder_Tasks();
 
   if (Transceiver_GetMode() == T_MODE_RESPONDER) {
+    RDMResponder_Tasks();
     RDMHandler_Tasks();
     SPIRGB_Tasks();
     Temperature_Tasks();
--- a/firmware/src/transceiver.c
+++ b/firmware/src/transceiver.c
@@ -99,9 +99,6 @@ typedef enum {
   STATE_ERROR = 100
 } TransceiverState;
 
-/*
- * @brief
- */
 typedef enum {
   OP_TX_ONLY = T_OP_TX_ONLY,
   OP_RDM_DUB = T_OP_RDM_DUB,
--- a/firmware/src/transceiver.c
+++ b/firmware/src/transceiver.c
@@ -46,9 +46,16 @@ enum { BUFFER_SIZE = DMX_FRAME_SIZE + 1u };
 // The number of buffers we maintain for overlapping I/O
 enum { NUMBER_OF_BUFFERS = 2};
 
+const int16_t TRANSCEIVER_NO_NOTIFICATION = -1;
+
+// Timing offsets
 static const uint16_t BREAK_FUDGE_FACTOR = 74u;
 static const uint16_t MARK_FUDGE_FACTOR = 217u;
 static const uint16_t RESPONSE_FUDGE_FACTOR = 24u;
+
+// The value of the test byte we send during the self test
+static const uint8_t SELF_TEST_VALUE = 0xa5;
+static const uint32_t SELF_TEST_TIMEOUT = 100;  // 10ms
 
 typedef enum {
   // Controller states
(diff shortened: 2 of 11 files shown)


firmware/src/app.c
```diff
@@ -156,9 +156,9 @@ void APP_Tasks(void) {
   USBTransport_Tasks();
   Transceiver_Tasks();
   USBConsole_Tasks();
-  RDMResponder_Tasks();
 
   if (Transceiver_GetMode() == T_MODE_RESPONDER) {
+    RDMResponder_Tasks();
```
peternewman: Haven't we seen this already?

firmware/src/transceiver.c
```diff
@@ -95,6 +108,7 @@ typedef enum {
   OP_RDM_BROADCAST = T_OP_RDM_BROADCAST,
   OP_RDM_WITH_RESPONSE = T_OP_RDM_WITH_RESPONSE,
   OP_RX = T_OP_RX,
+  OP_SELF_TEST = T_OP_SELF_TEST,
   OP_RDM_DUB_RESPONSE,  //!< No break
   OP_RDM_RESEPONSE  //!< With a break
 } InternalOperation;
```
peternewman: Can you Doxygen this enum, or remove the blank @brief so we know it's not been documented.

firmware/src/transceiver.c
```diff
@@ -1332,10 +1424,72 @@ void Transceiver_Tasks() {
       g_transceiver.data_index = 0u;
       g_transceiver.state = STATE_R_RX_PREPARE;
       break;
+
+    // Self Test States
+    case STATE_T_INITIALIZE:
+      PLIB_USART_TransmitterDisable(g_hw_settings.usart);
```
peternewman: Does this need to check the mode first too?

firmware/src/transceiver.h
```diff
@@ -23,26 +23,48 @@
...
+ * receives the same token and any operation-specific data. This allows the
+ * client to associate the callback with the original request that generated
+ * the event.
+ *
+ * If a TRANSCEIVER_NO_NOTIFICATION is used for the token, the callback will
+ * not be run.
+ *
+ * The transceiver can operate in various modes:
+ *  - DMX / RDM Controller
+ *  - DMX / RDM receiver
```
peternewman: SPaG consistency Receiver.

firmware/src/transceiver.h
```diff
@@ -254,11 +284,16 @@ void Transceiver_Initialize(const TransceiverHardwareSettings *settings,
 /**
  * @brief Change the operating mode of the transceiver.
  * @param mode the new operating mode.
+ * @param token to token to the TransceiverEventCallback when the mode change
```
peternewman: SPaG I'm not sure exactly what you're trying to say, all I know is it's not right. :)

firmware/src/transceiver.h
peternewman: Done

firmware/src/transceiver.h
peternewman: Possibly the second token should be pass?

peternewman:
We're happy to take the risk of a false result if two units already on the same DMX bus run self test at the same time, with one broken Tx?

peternewman:
Generally LGTM

firmware/src/transceiver.c
nomis52: We can only end up here from the SwitchMode() function.

firmware/src/transceiver.c
nomis52: Removed

firmware/src/app.c
nomis52: Yup, synced with master and now this diff has gone.

firmware/src/transceiver.h
nomis52: Fixed :)

nomis52:
The self-test is depends on the unit being disconnected from the DMX bus. I'll make a note of that somewhere.

firmware/src/transceiver.c
peternewman: Is that a no then @nomis52 ?

peternewman:
LGTM

firmware/src/transceiver.c
nomis52: Yeah, we don't need it.

PR #239, opened 2015-10-30
Add the simulator so we can get some coverage of the transceiver module.

nomis52:

PR #240, opened 2015-11-03
More tests

nomis52:
Depends on #239


peternewman:
LGTM, probably remove the commented printfs.

PR #244, opened 2015-11-04
Add a tranceiver test for responding to an RDM command

nomis52:


peternewman:
So given this outstanding coverage level, do we now make that required too, so we don't slip back down? Can admin currently override it?

firmware/src/transceiver.c
```diff
@@ -364,6 +364,15 @@ bool UART_RXBytes() {
 // Memory Buffer Management
 // ----------------------------------------------------------------------------
 
+
+/*
+ * @brief Return the number of free buffers.
+ * This is exposed for testing purposes.
```
peternewman: I think this needs a blank line so it's not part of the @brief?

firmware/src/transceiver.c
```diff
@@ -906,16 +926,11 @@ void __ISR(AS_USART_ISR_VECTOR(TRANSCEIVER_UART), ipl6AUTO)
...
-        // RX buffer is full.
-        PLIB_TMR_Stop(g_hw_settings.timer_module_id);
-        SYS_INT_SourceDisable(g_hw_settings.usart_rx_source);
-        SYS_INT_SourceDisable(g_hw_settings.usart_error_source);
-        PLIB_USART_ReceiverDisable(g_hw_settings.usart);
-        ResetToMark();
-        g_transceiver.result = T_RESULT_RX_INVALID;
-        g_transceiver.state = STATE_C_COMPLETE;
-      }
+      // It's impossible to overflow the buffer here, because each byte is 44uS
```
peternewman: Is there any reason to remove the code though, in case someone changes one of the variables? Or at least add a note where rdm_dub_response_limit is set.

firmware/src/transceiver.c
```diff
@@ -545,9 +557,12 @@ static inline void PrepareRDMResponse() {
   if (g_timing_settings.rdm_responder_jitter) {
     jitter = Random_PseudoGet() % g_timing_settings.rdm_responder_jitter;
   }
+  // It's important to stop the timer before changing the period, see 14.3.11
```
peternewman: Is it worth a regex in our cpplint task or similar that ensures PLIB_TMR_Period16BitSet and others are always preceded by PLIB_TMR_Stop?

tests/tests/SimulatedTransceiverTest.cpp
```diff
@@ -207,6 +234,19 @@ class TransceiverTest : public testing::Test {
   }
 
   void TearDown() {
+    if (!::testing::Test::HasFatalFailure()) {
+      // Run for another 6ms to allow any final states to timeout
+      m_simulator.SetClockLimit(6000, false);
```
peternewman: Doesn't this extra timeout want to be fatal too?

tests/tests/SimulatedTransceiverTest.cpp
```diff
@@ -642,9 +797,46 @@ TEST_F(TransceiverTest, rxInterSlotDelay) {
...
+      Run(EventIs(token, T_OP_RX, T_RESULT_RX_FRAME_TIMEOUT,
+                  arraysize(expected_frame))))
+    .WillOnce(AppendTo(&rx_data));
+
+  m_generator.SetStopOnComplete(true);
+  m_generator.AddDelay(100);
+  m_generator.AddBreak(176);
+  m_generator.AddMark(12);
+
+  // We can have up to 2.1ms between DMX slots.
```
peternewman: Why does the comment say DMX, but you're sending RDM?

peternewman:
Generally LGTM

firmware/src/transceiver.c
nomis52: Done

firmware/src/transceiver.c
nomis52: To fill the buffer you'd have to go way out of spec, even further than what we allow for testing. 

Added a comment in the setter.

firmware/src/transceiver.c
nomis52: This is done in the tests now, which is a much more robust way than trying to parse the code.

tests/tests/SimulatedTransceiverTest.cpp
nomis52: No, this one just lets the state machine settle.

tests/tests/SimulatedTransceiverTest.cpp
nomis52: Fixed the comment.

nomis52:
I don't want to add a hard limit on coverage.

firmware/src/transceiver.c
peternewman: They complain if the timer isn't stopped when it's changed? Awesome.

tests/tests/SimulatedTransceiverTest.cpp
peternewman: Okay, I thought there might be a risk of it getting stuck in another loop.

firmware/src/transceiver.c
nomis52: Not just complain, the fail!

peternewman:
Okay, fair enough.

peternewman:
LGTM

PR #245, opened 2015-11-04
Add some more tests

nomis52:


firmware/src/transceiver.c
```diff
@@ -872,8 +876,6 @@ void __ISR(AS_USART_ISR_VECTOR(TRANSCEIVER_UART), ipl6AUTO)
           SYS_INT_SourceStatusClear(g_hw_settings.input_capture_source);
           SYS_INT_SourceEnable(g_hw_settings.input_capture_source);
 
-          // TODO(simon) I think we can remove this because its done in the IC
```
peternewman: That's a no to removing then?

nomis52: Correct.

firmware/src/transceiver.c
```diff
@@ -1221,15 +1233,53 @@ void Transceiver_Tasks() {
...
+        return;
+      }
+      SYS_INT_SourceEnable(g_hw_settings.input_capture_source);
       break;
 
     case STATE_C_RX_DATA:
-      // TODO(simon): handle the timeout case here.
-      // It's not a static timeout, rather it varies with the slot count.
-      // PLIB_TMR_Stop(g_hw_settings.timer_module_id);
+      // There is not hard timeout on RDM responses, instead it depends on the
```
peternewman: SPaG not

firmware/src/transceiver.c
```diff
@@ -1221,15 +1233,53 @@ void Transceiver_Tasks() {
...
+      // number of slots sent, see Table 3.3.
```
peternewman: Is slot the right word when it's not DMX, surely they're just bytes?

firmware/src/transceiver.c
peternewman: Or no even, anyway fixed.

firmware/src/transceiver.c
peternewman: This isn't actually outdated @nomis52

peternewman:
Generally LGTM, just one comment left.

firmware/src/transceiver.c
nomis52: Yeah, even in RDM they are called slots.

peternewman: Okay. The outdated comment was because GitHub hid it when I edited the line above.

peternewman:
LGTM

PR #246, opened 2015-11-05
Add some more tests

nomis52:


firmware/src/dimmer_model.c
```diff
@@ -1315,7 +1315,7 @@ static const PIDDescriptor SUBDEVICE_PID_DESCRIPTORS[] = {
   {PID_CLEAR_STATUS_ID, (PIDCommandHandler) NULL, 0u,
     DimmerModel_ClearStatusId},
   {PID_SUB_DEVICE_STATUS_REPORT_THRESHOLD,
-    DimmerModel_GetSubDeviceReportingThreshold, 1u,
```
peternewman: How did you spot this, it looks like we don't have any tests for it in the RDM tests?

tests/tests/DimmerModelTest.cpp
```diff
@@ -139,3 +151,593 @@ TEST_F(DimmerModelTest, dmxBlockAddress) {
   EXPECT_THAT(ArrayTuple(g_rdm_buffer, size), ResponseIs(response.get()));
 }
 
+TEST_F(DimmerModelTest, statusMessage) {
+  uint8_t status_type = 0x02;
```
peternewman: I take it you don't want to pull in the OLA includes here?

tests/tests/DimmerModelTest.cpp
```diff
@@ -139,3 +151,593 @@ TEST_F(DimmerModelTest, dmxBlockAddress) {
...
+  size = InvokeRDMHandler(request.get());
+  EXPECT_THAT(ArrayTuple(g_rdm_buffer, size), ResponseIs(response.get()));
+}
+
+TEST_F(DimmerModelTest, modulationFrequencyDescription) {
+  uint8_t setting = 1;
+  unique_ptr<RDMRequest> request = BuildSubDeviceGetRequest(
+      PID_MODULATION_FREQUENCY_DESCRIPTION, 1, &setting, sizeof(setting));
+
+  const uint8_t expected_response[] = "\001\000\000\000250Hz";
```
peternewman: Shouldn't this be mod freq id, mod freq (32 bits, i.e. add another byte with 250 as it's value), description?

peternewman:
Generally LGTM

firmware/src/dimmer_model.c
nomis52: From the unit tests :)

Right now the RDM responder tests don't have good sub device coverage.

tests/tests/DimmerModelTest.cpp
nomis52: In array form it's

{ 1, 0, 0, 0, 50, '5', '0',  'H', 'z' } 

50 is ASCII '2'

tests/tests/DimmerModelTest.cpp
nomis52: Not the OLA ones, but I switched to using STATUS_ADVISORY from rdm.h

firmware/src/dimmer_model.c
peternewman: Ah of course, because this would flag the normal request as incorrect. I was confused as you said you weren't doing WithData type requests before.

tests/tests/DimmerModelTest.cpp
peternewman: Can you be consistent then please, and use \050 to avoid this confusion.

nomis52: Done, it's \062

peternewman: Ah, didn't realise it was octal. Perhaps changing it all to hex would be less confusing.

PR #248, opened 2015-11-05
Add more tests

nomis52:


tests/tests/RDMResponderTest.cpp
```diff
@@ -95,6 +95,42 @@ int ClearSensors(const RDMHeader *header,
...
+    PREFIX_DECI
+  }
+};
+
+SensorData g_sensors[NUMBER_OF_SENSORS];
+
+/*
+const char PIXEL_TYPE_STRING[] = "Pixel Type";
+
+const ParameterDescription PIXEL_TYPE_DESCRIPTION = {
```
peternewman: For a future test I assume?

nomis52: Yup, I still have some more tests to add for this PR. Got to get to  > 90% .

tests/tests/RDMResponderTest.cpp
```diff
@@ -138,11 +174,41 @@ class RDMResponderTest : public testing::Test {
...
+      g_sensors[i].should_nack = false;
+    }
+  }
+
  protected:
   UID m_controller_uid;
   UID m_our_uid;
   StrictMock<MockPIDHandler> m_pid_handler;
 
+  unique_ptr<RDMRequest> BuildGetRequest(uint16_t pid,
```
peternewman: Don't we have these? What were you using before?

peternewman:
LGTM

tests/tests/RDMResponderTest.cpp
nomis52: We have similar ones in the ModelTest, but this is for the RDMResponder test, which isn't a model.

nomis52:
Over to you @peternewman

peternewman:
Generally LGTM

nomis52:
Fixed. The numbers are DMX footprint and Slot Count. I flipped it around regardless.

tests/tests/RDMResponderTest.cpp
```diff
@@ -95,6 +95,100 @@ int ClearSensors(const RDMHeader *header,
   return 0;
 }
 
+// Sensors
+enum { NUMBER_OF_SENSORS = 2 };
+
+const char SENSOR_NAME1[] = "Temperature";
+const char SENSOR_NAME2[] = "Accel";
```
peternewman: We could make this Acceleration and it would still be the same length as sensor 1. Although if I'm reading this, it's only used in the tests, it's not live in Ja Rule?

peternewman:
LGTM

PR #249, opened 2015-11-10
Add a response time fudge factor.

nomis52:
The ISR runs about 1.5us after the UART event. Experiments point to 13 as the
best value. From my tests we're accurate to within 0.2 of a microsecond from
the value recorded with the Logic.


peternewman:
LGTM

PR #253, opened 2015-11-16
Initial checkin of the flash code.

nomis52:
This writes the first two bytes to flash. I can't yet get the hardware based
end-of-write detection working.


nomis52:
I found the problem. You need to drive CE low after the AAI instruction. I need to think about how this will interact with us multiplexing the pixel control over the same SPI port.

PR #264, opened 2016-03-07
Add a check to ensure we're cpplinting everything

peternewman:
Given the custom lint configurations we run


peternewman:
I've not tested the user_manual/pid_gen/pid_gen.cpp change, but I assume it should be fine.

peternewman:
This is ready for review now @nomis52 .

PR #270, opened 2016-11-12
Hardcode the Coverity compiler version, as the variable isn't set at …

peternewman:
…the right time


peternewman:
Another retrospective @nomis52 .

PR #275, opened 2017-01-10
Fix some Coverity issues

peternewman:


peternewman:
These event tags actually work, so I should be able to tidy up some OLA ones too.

PR #282, opened 2018-02-22
Try and update to the new path for gtest

peternewman:


peternewman:
Take two @nomis52 , now I've actually fixed it!

PR #284, opened 2018-05-31
Add Codespell, closes #328

peternewman:


peternewman:
Can I have a re-review please @nomis52 . Just fixed the final one having checked it was spelt correctly in the DFU spec and just wrong for us...

